PR #3, opened 2015-03-23
Put in example of how to enter a private key.

mikegehard:
This will make it easier for newbies to figure this out.


vito:
thanks!

PR #10, opened 2015-06-02
provide error message when the private key has a passphrase.

ryfow:
I tried to use a private key that was passphrase protected. The git-resource failed but there wasn't an indication as to what went wrong. This commit prints a message to standard error that says what went wrong.

I'm really new to concourse, so I haven't gotten the whole build/test process figured out yet. I might not get much more time to work on this, so I thought I'd send you a pull request to find out what tests you want written.

### DIFF
--- a/assets/askpass.sh
+++ b/assets/askpass.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+echo "Private keys with passphrases are not supported." >&2
+exit 1
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -11,7 +11,7 @@ load_pubkey() {
     eval $(ssh-agent) >/dev/null 2>&1
     trap "kill $SSH_AGENT_PID" 0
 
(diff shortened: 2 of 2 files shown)


vito:
Nice, thanks; this has bitten a few people.

The tests for the git resource run entirely within `docker build`, so that should be your feedback loop. I think a reasonable test could just be to generate a keypair with `ssh-keygen` with a password, and try to use that generated key when running `out`.

The tests are pretty high-level and are written in bash so that we can easily run in the container, to ensure it's the exact same environment that we run in production. This is important since the environment in the container is pretty stripped down (to save on space).

See https://github.com/concourse/git-resource/blob/master/test/put.sh

Thanks again!

vito:
Ended up adding the test myself since I'd like to get this change in. Thanks again!

ryfow:
Glad it was useful. Sorry that I didn't get back to it.

Issue #14, opened 2015-08-05
Support shallow cloning


oppegard:
For large repos with years of history, it'd be useful to support the `--depth` option when cloning a repo.

xoebus:
Nice. Let's move continuing discussion on to the pull request (#15).

PR #15, opened 2015-08-05
Support shallow git clones.

oppegard:
This PR will address #14. I've tested the resource with the new param set to `50`, and not specified at all. This reduces our repo clones from ~60 seconds to ~3 seconds. :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ Submodules are initialized and updated recursively.
 
 #### Parameters
 
+* `depth`: *Optional.* If a positive integer is given, *shallow* clone the
+  repository using the `--depth` option.
+
+
 * `fetch`: *Optional.* Additional branches to fetch so that they're available
   to the build.
 
--- a/assets/in
+++ b/assets/in
@@ -27,6 +27,7 @@ load_pubkey $payload
 uri=$(jq -r '.source.uri // ""' < $payload)
 branch=$(jq -r '.source.branch // ""' < $payload)
 ref=$(jq -r '.version.ref // "HEAD"' < $payload)
+depth=$(jq -r '(.params.depth // 0)' < $payload)
 fetch=$(jq -r '(.params.fetch // [])[]' < $payload)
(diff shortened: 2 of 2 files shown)


xoebus:
Great, thanks for this!

Is there ever a case where you don't want to set it between either everything or 1? Maybe this could be simplified down to `shallow: true/false`.

oppegard:
I was hoping for a simpler boolean value as well. At first I tried setting it to 1, but triggering a prior build failed because the version of the git-resource at that time wasn't cloned. Changing it to 50 addressed the issue.

Another option I considered was adding a generic `clone_extra_args` param, where I could specify `--depth 50` as the value. This would increase flexibility for the resource, and not require adding a new param for every option we'd like to pass on to `git clone`.

vito:
@oppegard In general we avoid adding those kinds of open-ended features ('extra flags') as it couples us to whatever tool we're using underneath and potentially breaks when we change versions of the `git` cli. I think higher-level `source`/`params` makes more sense.

We initially supported `--depth` by default but ran into issues with it because we don't know how far back we need to clone so that we can `git checkout` the requested sha (so constrained inputs would often fail). Letting the user specify the depth is one way to get around that but it feels like kind of a trap. What'd be really nice is if Git supported a way to clone starting at a particular sha.

tracker-common:
Is that a blocker to merge this feature? I'm willing to deal with the pain of failed builds because I didn't specify a large enough depth. Until concourse volumes happen, I feel like a bad internet citizen cloning a 1.8GB repo over and over and over throughout the day.

oppegard:
Is that a blocker to merge this feature? I'm willing to deal with the pain of failed builds because I didn't specify a large enough depth. Until concourse volumes happen, I feel like a bad internet citizen cloning a 1.8GB repo over and over and over throughout the day. :cry:

vito:
Yeah fair. I think this is the best way to do it atm. Merging!

oppegard:
Thanks!

PR #17, opened 2015-10-15
Add option git_ignore_ssl to disable git SSL verification

jochenehret:
* sets environment variable GIT_SSL_NO_VERIFY to true
* works also for git submodules
* using env variable (instead of git config --global) has no side-effects
* useful if git repository uses a self-signed certificate


momchil-sap:
+1, this would be great!

devnal:
+1

vito:
Sorry for the delay on this. Would you mind renaming the config to `skip_ssl_verification`?

xoebus:
Thanks!

PR #20, opened 2015-11-18
Adds ability to only push tags

### LINKED ISSUE #19: tags-only push
the git resource does not seem to be able to tag a repo at the sha provided to a build, if there have been commits after that sha on the branch.

when we try to do this, we get errors on push if we don't specify rebase, and it tags the wrong sha if we do specify rebase.

a tags-only push would solve this problem for us

zachgersh:
This may be a slightly naive approach to this.  Are there other edge cases here?  Can make any other changes required.

Fixes #19 

Signed-off-by: Angela Chin <achin@pivotal.io>


zachgersh:
@xoebus - comments here?

vito:
lgtm, thanks!

PR #26, opened 2016-01-14
Install git lfs in resource image

cjcjameson:
- This will add ~3 MB to the image, we estimate


vito:
What's the externally visible delta? It'll now fetch your files for you on `pull` or something/

cjcjameson:
No external-facing delta for a repo that has no files tracked with git-lfs
The external-facing delta for a repo that does have files under git-lfs is that:
- Without git-lfs installed, a `git pull` gets you a pointer to the file
- With git-lfs installed, the large file pointer is smudged into an actual copy of the file at `fetch` time

Additional console output would provide evidence of the latter case happening. Pushing will also show a similar difference.

We don't think trying to push into lfs without lfs installed would be a reasonable workflow

vito:
Cool, sounds good to me. Will have a look after 0.71.0 (trying to not introduce more deltas for that one as we haven't released in a while). Thanks!

vito:
Just remembered this - will merge it in now. Thanks!

PR #31, opened 2016-02-03
Selective triggering based on a tag pattern

Jonty:
This adds the ability to specify a `trigger_tag_filter` that will only trigger pipelines when a matching tag is pushed to the repository. It will work with both lightweight and annotated tags, however it is recommended that you use annotated tags when pushing to public repositories as they have time and author metadata.

e.g. `trigger_tag_filter: *production` would match a semver tag of `0.1.0-production` and trigger the pipeline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
 * `skip_ssl_verification`: *Optional.* Skips git ssl verification by exporting
   `GIT_SSL_NO_VERIFY=true`.
 
+* `tag_filter`: *Optional*. If specified the resource will only `trigger` when
+  a tag matching the specified expression is pushed to the repository. Patterns
+  are [glob(7)](http://man7.org/linux/man-pages/man7/glob.7.html) compatible
+  (as in, bash compatible).
+
 ### Example
 
 Resource configuration for a private repo:
--- a/assets/check
+++ b/assets/check
@@ -22,6 +22,7 @@ uri=$(jq -r '.source.uri // ""' < $payload)
 branch=$(jq -r '.source.branch // ""' < $payload)
 paths="$(jq -r '(.source.paths // ["."])[]' < $payload)" # those "'s are important
 ignore_paths="$(jq -r '":!" + (.source.ignore_paths // [])[]' < $payload)" # these ones too
+tag_filter=$(jq -r '.source.tag_filter // ""' < $payload)
 ref=$(jq -r '.version.ref // ""' < $payload)
 
 destination=$TMPDIR/git-resource-repo-cache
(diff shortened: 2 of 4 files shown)


vito:
Functionality looks good but I think the name `trigger_tag_filter` is confusing. Resources don't determine what triggers, as much as they determine what versions they discover. How about just `tag_filter`?

Jonty:
Hah, I actually called it that originally but thought you'd want it to be more explicit. I've changed it back.

I should note that the tests added in this PR depend on https://github.com/concourse/git-resource/pull/32 as they require the helper method to generate annotated tags.

README.md
```diff
@@ -29,6 +29,11 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
 * `skip_ssl_verification`: *Optional.* Skips git ssl verification by exporting
   `GIT_SSL_NO_VERIFY=true`.
 
+* `tag_filter`: *Optional*. If specified the resource will only `trigger` when
```
vito: Description should instead say "resource will only detect commits with tags matching the filter" or some such (not trigger). I'll just fix that up after merging though.

vito:
Looks good, thanks!

vito:
@Jonty Having given this another look, I noticed a few things that I probably should've noticed beforehand. They're not necessarily issues, just wanted to call them out to make sure they're in line with your expectations:

1. `check` will only yield one version (the latest) - which is fine for the first `check` but if checking from e.g. `1.0.0-production` and there's a `1.1.0-production` and `1.2.0-production` it would be nice if it could find both, and not just `1.2.0-production`. Note that this doesn't mean semver despite my example, just that it should find all applicable tags since the current one, ordered by the commits that they correspond to.
2. The versions emitted will be something like `{"ref": "1.1.0-production"}`, so with the tag as the `ref` rather than the sha that the tag corresponds to. Not wrong, and probably more useful to show in the UI, I'm just a little surprised that it works (does it?), since `in` just does a `git clone` and then a `git checkout`, and I'm not sure if the tags will all be present after just cloning, for the `checkout` to work.

If you want to try out the resource in your deployment to make sure it works for your workflow, you can point your workers at the `concourse/git-resource` Docker image, since we've already merged it.

Jonty:
@vito No problem, glad you're keeping a close eye on things!

1. This was a conscious choice as bare git tags don't have a timestamp associated with them - while you can get the _last_ tag, you can't get the last few bare tags ordered reliably. I added annotated tag support in another request specifically to resolve this issue but decided that keeping compatibility with bare tags as well as annotated tags was probably the right thing to do. I note that the git maintainers will scold you for pushing bare tags so this is supporting bad behaviour - if you felt strongly about this I'd be happy to submit a new PR that only supports annotated tags and causes the `check` script to return all new tag versions.

2. Git deals with this absolutely fine, all tags are fetched as part of the initial clone - I considered changing the clone to misuse `--branch` and checkout just the tag, but worried that not having a complete copy of the repo would be unexpected for users. I thought it was kind of neat that the ui showed the tag though!

vito:
@Jonty 

1. Right, the tags themselves aren't comparable, but it could determine the ordering by comparing the shas that the tags point to. I guess that only works if both shas live on the same branch, but that's what the Git resource is meant for anyway (one branch). Come to think of it it should also probably ignore tags for commits living on other branches. To sum it up, I think we want something like "list all commits on branch (as it does today) -> find tags for each commit -> filter to commits + tags matching the pattern -> emit all remaining tags".

2. Cool, sounds fine then.

PR #32, opened 2016-02-03
Enable annotated tagging

Jonty:
This allows the resource to push annotated tags as well as lightweight tags, meaning they have metadata attached such as the author, timestamp, and a message. It is expected that users would use annotated tags as part of a release process.

e.g. passing `annotate: foo/bar/annotation_file` to a resource tag operation will result in an annotated tag being created with the contents of `foo/bar/annotation_file` as the annotation message.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -32,6 +32,7 @@ tag=$(jq -r '.params.tag // ""' < $payload)
 tag_prefix=$(jq -r '.params.tag_prefix // ""' < $payload)
 rebase=$(jq -r '.params.rebase // false' < $payload)
 only_tag=$(jq -r '.params.only_tag // false' < $payload)
+annotation_file=$(jq -r '.params.annotate // ""' < $payload)
 
 if [ -z "$uri" ]; then
   echo "invalid payload (missing uri)"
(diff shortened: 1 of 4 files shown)


assets/out
```diff
@@ -60,11 +61,16 @@ if [ -n "$tag" ]; then
   tag_name="$(cat $tag)"
 fi
 
+annotate=""
+if [ -n "$message" ]; then
+  annotate=" -a -m '$(cat $message)'"
```
vito: I'm afraid of interpolating messages like this; seems like it would break if their message had a `'` in it.

Looks like there's a `-F` flag for reading the message from a file, so this could become:

```sh
  annotate=" -a -F '$message'"
```

vito:
Just to confirm: `annotate` is a path to a file, and not a string containing the message as in your description of the PR. (I think the file makes more sense.)

Jonty:
Ugh, this is what you get for eagerly pushing upstream at the end of the day when ill.

I've corrected the PR comment and the commit message to reflect the file-based nature, and amended the annotation to use git's built-in file-reading capabilities. Good catch!

vito:
Cool, merging. Thanks!

Issue #5, opened 2015-04-11
in parameter - specify a tag based on a file


drnic:
I'd like to be able to `git checkout` my `cf-release` repo to a specific tag so I can access the correct `templates` folder.

mariash:
I believe this is implemented https://github.com/concourse/git-resource#source-configuration

fewknow:
@mariash I am looking for this same functionality but I am not seeing it on the resource.  I have tried `tag_filter` to clone a particular tag when my `- get` is executed for the repository.   I also do not see anything in the documentation for this resource that lets you filter on the "In".   Please advise on how to clone a particular tag.

PR #35, opened 2016-02-12
Allow pinning of the git resource ref

Jonty:
This allows users to specify the version of the git repository that should be checked out rather than following master. It is designed to be used in situations where a third-party git repository is part of a pipeline and you wish to have control over the version available, or where your pipeline determines the required version of a repository at runtime.

The `check` action gains a `pin_ref` parameter that expects a valid git reference, and will cause the resource to remain pinned at that reference. This should be used when you wish to lock build dependencies from your pipeline configuration, such as the version of cf-release.

The `in` action gains a `pin_ref_file` parameter that expects a path to a file containing a valid git reference, and will cause the resource to remain pinned at that reference. This should be used when the version of the repository required is determined by a step earlier in the pipeline.

Theoretically this should resolve https://github.com/concourse/git-resource/issues/5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
   [glob(7)](http://man7.org/linux/man-pages/man7/glob.7.html) compatible (as
   in, bash compatible).
 
+* `pin_ref`: *Optional.* Pins the git repository at the specified reference
+  instead of tracking master. Expects a valid git reference.
+
 ### Example
 
 Resource configuration for a private repo:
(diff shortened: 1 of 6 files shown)


vito:
How does this compare to https://github.com/concourse/concourse/issues/256? I would much rather implement a generic solution than have resources support various forms of pinning.

Jonty:
@vito If you want to genericise it, I presume you'd like to implement something like the following:

1) Resources have an optional `version` or `version_file` parameter that overrides the `ref` passed to the `in` script.
2) If a `version` or `version_file` parameter is specified, the check script is no longer run as it has no purpose.

Does that make sense? We may be interested in implementing this.

vito:
@Jonty I think it's better if resources don't have to know anything about this feature. They'd just have `in` called with a version and dutifully pull it down. `check` would continue to be called since it's an individual `get` step that would lock it down, not the entire resource.

Jonty:
@vito Right, seems sensible. I'll close this PR and see if I can drum up support for a proper implementation.

PR #37, opened 2016-03-03
Use updated ca certificates (from Mozilla via haxx.se)

tom-meyer:
We found that the docker image for git-resource did not have updated CA certificates. Adding a fresh list of certs solved our problem.

Thanks!

-Tom Meyer

p.s. I am a new-ish member of the Hawq Toolsmiths team in PA. I can be reached at tmeyer@pivotal.io


vito:
Thanks for this. We should probably just bump our Buildroot dependency though:

https://github.com/concourse/core-busyboxplus/blob/master/git/tarmaker/Dockerfile#L24

They seem to be pretty good about keeping their `ca-certifications` package [up to date](https://github.com/buildroot/buildroot/commit/30e991a0826f862c0b9b6e61fc498ca7358ef1d8), we just haven't bumped in a while.

vito:
We've bumped the version of Buildroot to `2016.02` (from `2015.02`), which should bring in new certs. If you want to give it a try the image is `concourse/git-resource` on the public Docker Hub.

tom-meyer:
Super, thanks! We'll give the new image a whirl.

PR #38, opened 2016-03-03
Disable git lfs for `check`

monkeyherder:
Previously, git-lfs was downloading all lfs artifacts as part of `check`. For large repositories, this could add hours onto the time to run a check script.


monkeyherder:
Ran into a problem with this on further testing. Will look into it and re-submit.

PR #40, opened 2016-03-05
Better git lfs handling

monkeyherder:
Previously, we were using the built in git-lfs functionality to download files at clone time. However, that's incredibly slow for repos with large numbers of git-lfs managed files, and is error prone (if a single file fails to download, the clone fails). Switch to using the deferred downloading of files with a `git lfs fetch` and then `git lfs checkout` after cloning, which includes retry logic and does batch downloading, reducing the download time by a factor of 10 or better for repos with large numbers of `git-lfs` managed files.

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -53,6 +53,8 @@ git clone --single-branch $depthflag $uri $branchflag $destination
 cd $destination
 
 git checkout -q $ref
+git lfs fetch
+git lfs checkout
 git log -1 --oneline
 git clean --force --force -d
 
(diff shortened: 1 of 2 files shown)


vito:
Makes sense, thanks!

Minor thing I want to confirm before merging: the `lfs` commands silently do nothing if you're not using LFS, right? I just don't want LFS-related things showing up in the clone output of everyone's git resources.

monkeyherder:
Yeah, both `git lfs fetch` and `git lfs checkout` are no-ops for repos without `git-lfs` managed files.

PR #41, opened 2016-04-01
Update common.sh to allow DSA keys.

avongluck-r1soft:


vito:
Given how scary [this](https://www.gentoo.org/support/news-items/2015-08-13-openssh-weak-keys.html) sounds I think we may just leave it as-is. Thanks though.

PR #44, opened 2016-04-12
username and password auth for HTTP(S) added

MatthiasWinzeler:
We're sitting behind a corporate firewall that only allows HTTPS access to our git repositories.
In such scenarios, public key - which is only possible using ssh - is not possible and username and password auth needs to be used.

This PR introduces two options, `username` and `password` which can be used for authenticating with HTTP(S).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,12 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
       -----END RSA PRIVATE KEY-----
     ```
 
+* `username`: *Optional.* Username for HTTP(S) auth when pulling/pushing.
+  This is needed when only HTTP/HTTPS protocol for git is available (which does not support private key auth)
+  and auth is required.
+
+* `password`: *Optional.* Password for HTTP(S) auth when pulling/pushing.
+
 * `paths`: *Optional.* If specified (as a list of glob patterns), only changes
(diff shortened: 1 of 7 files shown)


mariash:
Hi @MatthiasWinzeler, what is the reason to use .netrc approach. Can you just put username/password in url?

srbry:
Might not be the same reason as I am seeing but I want to use secret replacement so i dont save my passwords with my pipeline, doing this has the following error:

secrets file:
```
username: test_user
password: test_password
```
pipeline.yml  `uri: https://{{username}}:{{password}}@github.com/example`

Expected concourse config: `https://test_user:test_password@github.com/example`

Actual concourse config: `https://"test_user":"test_password"@github.com/example`

As i work around I am replacing my entire uri, it would be nice if i could just pass in a username and password as appropriate

mariash:
I see, thanks for explanation.

MatthiasWinzeler:
Hi @mariash
I used the `.netrc` solution because it uses the credentials for submodules also. Otherwise, we would have to put username and password into the submodule urls inside the repo (I'm not sure that's even possible).

o-orand:
Hi,
It would be great to have this PR merged :smile:

Pirolf:
We have merged the pr :)

PR #46, opened 2016-04-12
get: returns the tag name if tag_filter defined

### LINKED ISSUE #45: When using tag_filter, the version reported by 'in' should be the tag name
When using `tag_filter`, the `check` command [reports the tag name instead of the hash in the version list](https://github.com/concourse/git-resource/blob/d952cdb40964b8f67e1ec023bc2907ab19dc0c88/assets/check#L56-L59)

But then the `in` command returns the [hash not the tag](https://github.com/concourse/git-resource/blob/d952cdb40964b8f67e1ec023bc2907ab19dc0c88/assets/in#L74)

I think it should report the tag if available.

keymon:
When using tag_filter, the check step will return a list of valid
tags as versions. In order to keep consistency, the get command
should also return the tag if the tag_filter is defined.

This implementation will try to check if $ref is a tag, if not use
the latest tag, if not use the hash reference.

Fixes #45

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -27,6 +27,7 @@ configure_git_ssl_verification $payload
 
 uri=$(jq -r '.source.uri // ""' < $payload)
 branch=$(jq -r '.source.branch // ""' < $payload)
+tag_filter=$(jq -r '.source.tag_filter // ""' < $payload)
 ref=$(jq -r '.version.ref // "HEAD"' < $payload)
 depth=$(jq -r '(.params.depth // 0)' < $payload)
 fetch=$(jq -r '(.params.fetch // [])[]' < $payload)
@@ -72,7 +73,13 @@ for branch in $fetch; do
   git branch $branch FETCH_HEAD
 done
 
+return_ref=""
+if [ -n "$tag_filter" ]; then
+  return_ref=$(git tag -l --points-at HEAD --sort='version:refname' $tag_filter | tail -n 1)
+fi
+[ -n "$return_ref" ] || return_ref=$(git rev-parse HEAD)
+
 jq -n "{
-  version: {ref: $(git rev-parse HEAD | jq -R .)},
+  version: {ref: $(echo "$return_ref" | jq -R .)},
   metadata: $(git_metadata)
 }" >&3
--- a/test/get.sh
+++ b/test/get.sh
@@ -135,6 +135,66 @@ it_can_get_and_set_git_config() {
   test "$(git config --global credential.helper)" == '!true long command with variables $@'
 }
 
+it_returns_ref_when_using_tag_filter() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  local ref2=$(make_annotated_tag $repo "1.0-staging" "a tag")
+  local ref2=$(make_commit $repo)
+  get_uri_with_tag_filter $repo "*-staging" $TMPDIR/destination | jq -e "
+    .version == {ref: $(echo $ref2 | jq -R .)}
+  "
+}
+
+it_returns_tag_when_using_tag_filter() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  local ref2=$(make_annotated_tag $repo "1.0-staging" "a tag")
+  local ref3=$(make_commit $repo)
+  local ref4=$(make_annotated_tag $repo "1.1-staging" "another tag")
+  get_uri_with_tag_filter $repo "*-staging" $TMPDIR/destination | jq -e "
+    .version == {ref: $(echo $ref4 | jq -R .)}
+  "
+}
+
+it_returns_tag_at_tag_when_using_tag_filter() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  local ref2=$(make_annotated_tag $repo "1.0-staging" "other tag")
+  local ref3=$(make_annotated_tag $repo "0.9-production" "a tag")
+  local ref4=$(make_commit $repo)
+  local ref5=$(make_annotated_tag $repo "1.1-staging" "another tag")
+  local ref6=$(make_commit $repo)
+  get_uri_at_ref_with_tag_filter $repo $ref2 "*-staging" $TMPDIR/destination | jq -e "
+    .version == {ref: $(echo $ref2 | jq -R .)}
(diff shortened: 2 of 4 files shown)


keymon:
Other alternative implementation to this is simply output the `$ref` directly. It makes sense, as the in command will only success if `$ref` exists and won't change HEAD at all, so it will always report the same `$ref` than the one specified input request.

```diff
diff --git a/assets/in b/assets/in
index 1ba4378..485d475 100755
--- a/assets/in
+++ b/assets/in
@@ -73,6 +73,6 @@ for branch in $fetch; do
 done
 
 jq -n "{
-  version: {ref: $(git rev-parse HEAD | jq -R .)},
+  version: {ref: $(echo $ref| jq -R .)},
   metadata: $(git_metadata)
 }" >&3
```

If you like more that other approach, let me know and I will change it. The test can be the same.

keymon:
Can you please comment on this issue? It is even more problematic, because it can cause the pipeline being triggered with an old commit, which is undesirable.

vito:
Somehow unwatched this repo so didn't get updates; sorry! Dragged it in.

assets/in
```diff
@@ -72,7 +73,15 @@ for branch in $fetch; do
   git branch $branch FETCH_HEAD
 done
 
+return_ref=""
+if [ -n "$tag_filter" ]; then
+  return_ref=$(git tag -l --points-at HEAD "$ref")
+  [ -n "$return_ref" ] || return_ref=$(git tag -l --points-at HEAD | tail -n 1)
+fi
+[ -n "$return_ref" ] || return_ref=$(git rev-parse HEAD)
+echo $return_ref
```
mariash: is this echo is a leftover from debugging?

assets/in
```diff
@@ -27,6 +27,7 @@ configure_git_ssl_verification $payload
 
 uri=$(jq -r '.source.uri // ""' < $payload)
 branch=$(jq -r '.source.branch // ""' < $payload)
+tag_filter=$(jq -r '.source.tag_filter // ""' < $payload)
```
mariash: could you please update the README with the new get option?

assets/in
```diff
@@ -72,7 +73,15 @@ for branch in $fetch; do
   git branch $branch FETCH_HEAD
 done
 
+return_ref=""
+if [ -n "$tag_filter" ]; then
```
mariash: how is tag_filter used to match the tags? it looks like it is not being used at all.

test/get.sh
```diff
@@ -135,10 +135,25 @@ it_can_get_and_set_git_config() {
...
 }
 
+it_returns_tag_when_using_tag_filter() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  local ref2=$(make_annotated_tag $repo "1.0-staging" "a tag")
+  local ref3=$(make_commit $repo)
+  local ref4=$(make_annotated_tag $repo "1.0-production" "another tag")
+  local ref5=$(make_commit $repo)
+  get_uri_tag_with_tag_filter $repo $ref2 "*-staging" $TMPDIR/destination | jq -e "
```
mariash: This looks like it is going to work only because the ref2 is passed, not because there is a glob matching happening.

test/get.sh
```diff
@@ -135,10 +135,25 @@ it_can_get_and_set_git_config() {
...
+    .version == {ref: $(echo $ref2 | jq -R .)}
+  "
+}
+
```
mariash: could you please add more tests for edge cases introduced in PR? like when tags are not found or when ref is not provided and tag_filter is provided.

assets/in
keymon: Sorry, you are right I missed that one. The resource behaviour was not affected thx to redirect stdout to 3 fd.

assets/in
keymon: I don't think `tag_filter` should be documented as part of the get, as it is documented already in the [source documentation](https://github.com/concourse/git-resource/#source-configuration). 

Also, this is not a new option, it has been added in #31. In this PR I check if it has been defined in the source.

But it makes sense to update the documentation to mention that `in/get` will report back the tag as reference.

assets/in
keymon: In this PR we only check if `tag_filter` is defined to decide if we should report back a tag in the version or the git hash (Note that maybe this is not the ideal behaviour implementation as commented in https://github.com/concourse/git-resource/pull/46#issuecomment-208901148)

But the matching of `tag_filter` of happens in the `check` and it was implemented in #31

test/get.sh
keymon: As commented, the `tag_filter` logic has been implemented in #31 

In this test we check that: 
 * When ref is a tag (`1.0-staging` in that test)
 * And `tag_filter` is defined
 * Then the same tag `1.0-staging` is returned.

I refer again to the comment https://github.com/concourse/git-resource/pull/46#issuecomment-208901148, as maybe the simplest solution is always return the input ref passed to in, as the get operation should always checkout the same version.

test/get.sh
keymon: Sure, I will do so.

keymon:
But I would like to debate the behaviour in some cases. Shall we always return the same `ref` as described in https://github.com/concourse/git-resource/pull/46#issuecomment-208901148 or try to find out if `ref` has tags pointing to it which match `tag_release`. If there are several tags pointing to `ref`, which one shall we return?

I will update this PR later today.

assets/in
keymon: Finally I use tag filter in the in command as well, as the tag (or `tag_filter`) might after check is called.

assets/in
mariash: fair enough.

assets/in
mariash: we tested the command below, $tag_filter will only match exactly, it seems like it is not taking in glob patterns.

keymon: Just as a side note: The globing works and it is covered by the tests. But if the command is tried manually, be sure to quote the wildcards to avoid the shell expand or change the wildcard expression.

keymon:
After discussing this with @mariash, we agree that the best option is simply return the incoming `$ref` without changing it at all, so we leave the logic related to tags in the `check` command. 

I will close this PR for reference and open a new one.

PR #47, opened 2016-04-12
in/out: output branch and tags for current commit in metadata

### LINKED ISSUE #9: Add branch to metadata
It would be nice that checked out branch would be added to metadata.

keymon:
Add two extra fields to metadata:

 * It includes a 'tags' field in the metadata as a list, which might be useful
   for later consumption of the resource. If no tag is pointing to the current commit, prints an empty list.
 * It includes a branch field in the metadata, which might be useful for later
   consumption of the resource. If no branch is pointing to the current commit, uses the commit reference.

This also fixes #9

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -36,13 +36,20 @@ git_metadata() {
   local committer=$(git log -1 --format=format:%cn | jq -s -R .)
   local committer_date=$(git log -1 --format=format:%ci | jq -R .)
   local message=$(git log -1 --format=format:%B | jq -s -R .)
+  local branch=$(git rev-parse --abbrev-ref HEAD)
+  if [ "$branch" == "HEAD" ]; then
+    branch="$commit"
+  else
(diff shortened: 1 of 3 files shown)


vito:
This doesn't seem to be working - it just prints the commit sha instead of the branch it's on. Also it would be nice to omit the `tags` field entirely if there are none.

keymon:
I will investigate why it does return the hash and come back to you.

About the tags the current behaviour was intended and tested, but  +1 to not return the tags, I will do a PR to omit them.

keymon:
@vito What do you think about change it to report the list of branches which include a given commit, like `git branch --contains $ref`?

keymon:
I am preparing a PR to only return branch and tags if they have a value https://github.com/keymon/git-resource/commit/2551823afa76d6e3b57acaf582db4eba40b64e1b

I need to spend some more time to see why exactly it never returns the branch in the case you mention.

PR #48, opened 2016-04-12
Add support to configure git options in the resource configuration

keymon:
Add the optional field git_config as a list of pairs name:value to set the global options of the git command.

This enables the user to configure different options of git, like credential helpers, compression, etc.

One use case is using [AWS git repositories, AWS codecommit,](https://aws.amazon.com/codecommit/) using aws client [as a git credentials helper](https://docs.aws.amazon.com/codecommit/latest/userguide/setting-up-https-unixes.html).

With this patch the user can extend the git resource docker image to add AWS cli and set `git_config` to setup the required `credential.helper` and `credential.UseHttpPath`.


mariash:
thank you for PR, it looks great!

PR #50, opened 2016-04-16
using amd64 version of git-lfs instead of 386

fralalonde:
386 version of git-lfs dies with signal 5 (debug trap) on Amazon Linux. x64 version works as expected.


vito:
Oh, yeah weird, that should totally be x64. Thanks!

PR #52, opened 2016-05-03
get returns same ref requested by concourse

### LINKED ISSUE #45: When using tag_filter, the version reported by 'in' should be the tag name

keymon:
> Note: this PR replaces #46

The in command does checkout the code at a given ref, but it does not change
it. In consequence it should report the same ref as passed from concourse
as version to checkout.

We only return a different version in the case no version is specified
by concourse (first get).

For instance when using tag_filter, the check step will return a list of valid
tags as versions. In order to keep consistency, the get command
should also return the tag if the tag_filter is defined.

We include a test to verify that get reports the same version for
different ref (hash or tag) pointing to the same commit.

Fixes #45

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,11 @@ allows you to commit to your repository without triggering a new version.
 ### `in`: Clone the repository, at the given ref.
 
 Clones the repository to the destination, and locks it down to a given ref.
-Returns the resulting ref as the version.
+
+It will return the same ref as version, unless ref is `HEAD`, which
+will result in the hash after locking down as the version.
 
 Submodules are initialized and updated recursively.
-
 
 #### Parameters
 
--- a/assets/in
+++ b/assets/in
@@ -72,7 +72,13 @@ for branch in $fetch; do
   git branch $branch FETCH_HEAD
 done
 
+if [ "$ref" == "HEAD" ]; then
+  return_ref=$(git rev-parse HEAD)
+else
+  return_ref=$ref
+fi
+
 jq -n "{
-  version: {ref: $(git rev-parse HEAD | jq -R .)},
+  version: {ref: $(echo $return_ref | jq -R .)},
   metadata: $(git_metadata)
 }" >&3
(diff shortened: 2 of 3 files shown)


keymon:
I want to add that we could implement it differently by not returning any ref in the version of the json response when no version is sent by concourse. 

From the concourse documentation this should keep the version consistent after the execution of get.

Let me know if you prefer that.

README.md
```diff
@@ -102,10 +102,11 @@ allows you to commit to your repository without triggering a new version.
 ### `in`: Clone the repository, at the given ref.
 
 Clones the repository to the destination, and locks it down to a given ref.
-Returns the resulting ref as the version.
 
-Submodules are initialized and updated recursively.
+It will return the same ref as version, unless ref is `HEAD`, which
```
vito: When would `ref` ever be "HEAD"?

keymon: The only case is that no version is passed to `in`, which in that case it checkouts the latest commit of the branch, but indeed it is not expressed properly. Let me rephrase it.

keymon: Changed in #52

maryamklabib:
Hello,

I am using the `tag_filter` and the git resource is incorrectly showing me a tag (and commits) that do not conform to the regex I specified.
I have set the tag as `tag_filter: v1.2.*`
I hijacked onto the container and saw the following:
```
root@a15bc671-378c-46c8-57f8-22f67190a977:/tmp/build/c896b73c/p-pks-integrations# git tag --sort=v:refname | tail -n 2
v1.2.11-build.3
v1.3.0-build.1
```

I expected it to only go up to `v1.2.11-build.3` and indeed this explains my pipeline's unexpected behavior. Did I specify something incorrectly or is there a bug?

vito:
@maryamklabib Are you doing a `put` to the same resource?

PR #53, opened 2016-05-04
check returns commits in topological order

keymon:
When check returns all the commits from a given ref, currently it returns them ~~without a specific order~~ in the default chronological order, although the first commit is the given ref and the last HEAD.

But this might be confusing for the end user when looking at the git history of the repos and the version list stored in concourse, which was reported by the check command.

An alternate approach is to show the commits in topological order, as explained by --topo-order in git-log [1]:

> Show no parents before all of its children are shown, and avoid
> showing commits on multiple lines of history intermixed.
> for example, in a commit history like this:
> ```
>    ---1----2----4----7
>    \       \
>     3----5----6----8---
> ```
>
> where the numbers denote the order of commit timestamps, git rev-list and friends with --date-order show the commits in the timestamp order: 8 7 6 5 4 3 2 1.
> With --topo-order, they would show 8 6 5 3 7 4 2 1 (or 8 7 4 2 6 5 3 1); some older commits are shown before newer ones in order to avoid showing the commits from two parallel development track mixed together.

[1] https://git-scm.com/docs/git-log

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -67,6 +67,6 @@ if [ -n "$tag_filter" ]; then
 } | jq -R '.' | jq -s "map({ref: .})" >&3
 else
 {
-  git log --grep '\[ci skip\]' --invert-grep --format='%H' $log_range $paths_search
+  git log --grep '\[ci skip\]' --topo-order --invert-grep --format='%H' $log_range $paths_search
 } | jq -R '.' | jq -s "map({ref: .})" >&3
 fi
--- a/test/check.sh
+++ b/test/check.sh
@@ -309,6 +309,25 @@ it_can_check_and_set_git_config() {
   test "$(git config --global credential.helper)" == '!true long command with variables $@'
 }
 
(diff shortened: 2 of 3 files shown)


vito:
Interesting. So this mainly affects merging branches. Is this the ordering that GitHub's UI shows?

keymon:
Github [in the commits view](https://github.com/concourse/git-resource/commits/master) seems to do the default order, reverse chronological order.

vito:
I think it makes more sense to return in the default order - especially given that it's what most UIs show (GitHub, `git log`, etc.). If you feel strongly otherwise I'd be interested in hearing why, though.

keymon:
Fine for me.

To be sincere I was hoping to solve some situations where git resource returns the git commits in a different order than what I expect. But I am not able to reproduce it in a consistent way and I don't know if this fixes it. 

I will close this for the time being.

vito:
OK, thanks!

PR #54, opened 2016-05-06
in/get do process source.git_config

keymon:
The in and get command were not processing the optional field
git_config to the global options of the git command added in PR #48.

The execution changes the global config of git, but the
tests were not restoring the state between executions. Because
that the in/out command tests were passing cleanly.


keymon:
Hi, this PR complements an already merged PR #48, can you have a look?

PR #55, opened 2016-05-06
Not return branch or tags in metadata if empty and fix branch retrieval

keymon:
As per comment https://github.com/concourse/git-resource/pull/47#issuecomment-217249875 we implement:

  * Return the branch and tags metadata only if they are not empty.

    The git_metadata function has been refactored, using the operator `'. +' from `jq` to simplify the code when adding optional fields in the metadata response, like committer, branch or tags.

 * We fix the branch name retrieval, by using `git show-ref --heads` to get branch names.

   Using rev-parse of head will always return the commit hash when working in a detached HEAD, even if the HEAD ref is being also referenced by the branch. The in command works detached whenever a ref is passed to the in command.

Apart of the unit tests in this repo, this has been tested with this example pipeline:

```
resource_types:
- name: git2
  type: docker-image
  source:
    repository: keymon/git-resource

resources:
- name: concourse-develop
  type: git2
  source:
    uri: https://github.com/concourse/concourse.git
    branch: develop

- name: concourse-master
  type: git2
  source:
    uri: https://github.com/concourse/concourse.git
    branch: master

- name: concourse-tag
  type: git2
  source:
    uri: https://github.com/concourse/concourse.git
    tag_filter: v1.2.0

jobs:
- name: test
  plan:
  - get: concourse-develop
    params: {submodules: none}
  - get: concourse-master
    params: {submodules: none}
  - get: concourse-tag
    params: {submodules: none}

```
metadata_branch_tags

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,6 @@ Pushing local commits to the repo:
   params: {repository: some-other-repo}
 ```
 
-
 ## Behavior
 
 ### `check`: Check for new commits.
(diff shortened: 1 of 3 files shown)


vito:
Looks good to me, thanks!

PR #60, opened 2016-05-16
Allows [ci_skip] markings to be ignored

zachgersh:
Signed-off-by: Ryan Moran <rmoran@pivotal.io>


zachgersh:
btw - we have been consuming our docker hub version with this change and it works great.

PR #63, opened 2016-06-15
added git lfs fetch and checkout for checked out submodules

winterstonegames:
With this change submodules using git-lfs will also work. Without no lfs managed files will have correct content.

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -70,6 +70,7 @@ elif [ "$submodules" != "none" ]; then
     git submodule update --init $depthflag --recursive $submodule
   done
 fi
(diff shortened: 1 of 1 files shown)


vito:
Will this work recursively? Kind of an edge case, but it might just be a matter of passing `--recursive`.

Also will this have any impact (i.e. printing messages or failing) on folks not using git LFS?

knifhen:
It should not effect people without git lfs, since the script already does this for everybody..
```
git lfs fetch
git lfs checkout
```
only just for the main repo and not for submodules

vito:
Makes sense, thanks!

PR #69, opened 2016-07-14
Fetch tags when running check

aquavitae:
My use case is versions being managed by git tags. The scenario is that at the end of a build a new tag is pushed for the current githash with the version number. However, triggering another build with the same hash doesn't see the tag, assumes it doesn't exist and tries to push it again.  

I can think of other cases where one would always want the latest repo available, even if the build is run at a specific githash and I can't think of any scenario where pulling tags would cause a problem, but  if there is such a case then perhaps this could be paramatized?

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -34,7 +34,7 @@ destination=$TMPDIR/git-resource-repo-cache
 
 if [ -d $destination ]; then
   cd $destination
-  git fetch
+  git fetch --tags
   git reset --hard FETCH_HEAD
 else
   branchflag=""


vito:
I don't think this will fix your problem. The problem is that the checkout of the repo has been cached, and has tags present. I think what we should do long-term is remove *all* tags from the repo checked out by `get`, and force people to run `git fetch --tags`.

aquavitae:
Thanks for the response, but I don't follow. Why would this not fix it? How would removing all tags help? Where would the "git fetch --tags" live - in a task?

vito:
It sounded like your issue was that fetching the same SHA on separate occasions would have some tags present at first, and then tags would be added externally, and the new (or updated) tags would not be present in later fetches. This is due to resource caching, and won't be fixed: the resource should provide only the data it can safely cache (i.e. the sha itself and all history leading up to it). Tasks should be forced to do `git fetch --tags` if they want to interact with the tags in the repo, as those are separate (unversioned) state.

Adding `--tags` to the `/check` script won't change anything; `git` will already fetch tags for any refs it fetches. It also runs in a separate container with a separate filesystem than what ultimately gets pulled down.

aquavitae:
Ok, it makes sense from a design perspective, although the behaviour is rather unexpected - it is the same git resource pushing the new tag as the one pulling at the start of the build.  I'll close the PR though.

aquavitae:
Just an additional comment though - doing a `git fetch` in a task does break the concept of it being a pure function, but I guess that's the compromise you're making: "resources only manage versioned state" is more important that "tasks are pure functions".  FWIW I disagree with this compromise.

vito:
then there should either be a resource specifically for git tags, or the
git resource should idempotently push tags. I would like to see the latter,
as that's one principle of all resources (idempotency)

vito:
@aquavitae re-reading what you mentioned above that sounds like the real issue - tag pushing should be idempotent if it's pushing the same tag to the same sha. does that sound reasonable?

aquavitae:
Yes, that's exactly my issue!

vito:
created https://github.com/concourse/git-resource/issues/72 instead

PR #76, opened 2016-09-05
Add support for verifying commit GPG signatures

alext:
## What

Add support for verifying GPG signed commits.

This adds 3 new optional parameters to the resource:
 * `commit_verification_keys`: List of (armored) GPG public keys to import.
 * `commit_verification_key_ids`: List of GPG public keys import from remote servers.
 * `gpg_keyserver`: Optional, keyserver to download the keys from.
   Defaults to keys.gnupg.net

If `commit_verification_keys` or `commit_verification_key_ids` is defined, the resource will import all the specified public keys (downloading them from the public server gpg_keyserver as necessary) and verify that the the current commit is gpg signed by one of the listed keys. It will fail for any non-signed commit, invalid signatures or if signed by an unknown key.

This PR is dependent on https://github.com/concourse/buildroot-images/pull/3 being merged, and published to dockerhub.


vito:
Awesome! Really appreciate the amount of work that went into this, especially the testing and updating the README. After a quick skim it looks great.

Rather than switching out the image, however, I wonder if we can just update the existing image to set up GPG. Should be pretty easy.

The repo for doing that lives here:

https://github.com/concourse/buildroot-images

And the script for updating that is:

https://github.com/concourse/buildroot-images/blob/master/scripts/reconfigure

i.e. `./scripts/reconfigure 2016.08 git`.

That'll launch a terminal menu UI and somewhere in there should be a `gnupg` package to install.

alext:
Hi, thanks for the feedback.

I'll look into getting gpg working in the buildroot image. (I haven't gone down that route initially because I wanted to get feedback on whether this feature was likely to be accepted before investing the effort.)

alext:
I've been looking at implementing this on the buildroot image. The feature works fine, but getting the tests working was more difficult, and ended up deep in the differenced between GnuPG 2.0 and  2.1.  As a result, we've ended up with 2 different approaches to solve this:

## 1. Generating signed commits from the tests:

This can be seen here: https://github.com/concourse/git-resource/compare/master...alphagov:buildroot_without_fixture
This has the drawback that the tests will break if/when gnupg in the base image is upgraded to 2.1. It would be possible to extend this to check for the version of gnupg, and adapt the behaviour accordingly, but this won't deal with any future changes (who knows what changes gpg 2.2 will make)

## 2. Use a fixture repo that already contains a pre-generated signed commit.

This can be seen here: https://github.com/concourse/git-resource/compare/master...alphagov:buildroot_with_fixture
This will be more robust to changes in the gpg version, but does introduce a new approach to handling the tests, whereas the first approach is more in keeping with the exsiting patterns.

Do you have a preference for either of these approaches?

alext:
After thinking about this a bit more, I think I prefer the fixture-based approach - it's less likely to break in future due to changes in GPG. I've therefore updated this PR to use that. I'll keep the other branch around in case you prefer that approach.

vito:
I would prefer the first approach, if it's low-enough maintenance to get it working again. Do you know what kind of breakage to expect? How painful was 2.0 -> 2.1?

vito:
Woops, GitHub went single-page-app on me and didn't show your reply there. I don't feel too strongly either way to be honest.

alext:
> Do you know what kind of breakage to expect? How painful was 2.0 -> 2.1?

Basically, the approach to providing a passphrase in a headless manner has changed between the 2 versions. 2.1 removes the `--passphrase-file`, `--passphrase-fd` etc options. We therefore had to use the (newly introduced) loopback pinentry mode combined with seeding the passphrase into gpg-agent.

It's probably easiest to see the differences by comparing `test/helpers.sh` from the 2 implementations:

* for gpg 2.0: https://github.com/alphagov/paas-git-resource/blob/buildroot_without_fixture/test/helpers.sh#L124
* for gpg 2.1: https://github.com/alphagov/paas-git-resource/blob/gpg_verification_alpine/test/helpers.sh#L125

The concern I had was whether it was reasonable to expect whoever has to handle the base container upgrade to figure out these changes.

vito:
Gotcha. I'm OK with the fixture; it looks pretty small. I'll go ahead and merge this then - should be out in 2.1 today.

Thanks again!

alext:
Thanks for merging. I don't think this will work until https://github.com/concourse/buildroot-images/pull/3 is merged and published though.

vito:
@alext Merged it, but the build still seems to have failed:

https://ci.concourse.ci/teams/main/pipelines/resources/jobs/git-resource/builds/39

vito:
Fixed it up. Turned the readme into a script to run to generate it and regenerated it. After that it worked. Maybe some file was missing from your checking? Either way, build's green now; should be in 2.1.

alext:
Wow, sorry about that. I'd run a `git gc` on the fixture repo to make it compact, but that meant that the `refs` directory was empty, so git ignored it. It would seem that this directory is required, even if empty.

Thanks for fixing.

Issue #79, opened 2016-09-12
Ability to disable git LFS


philk:
We use git lfs for some development dependencies (~2GB of them) but we don't need them on CI (and it slows down the builds enormously).

It looks like it should be as easy as adding a param like `disable_git_lfs` and adding an `if` to the `inputs` script. Does that sound reasonable if I make a PR?

vito:
fixed by https://github.com/concourse/git-resource/pull/80

PR #80, opened 2016-09-15
Add param disable_git_lfs to assets/in

SHyx0rmZ:
This adds a new param called `use_lfs` to `assets/in`, which can be used to `get` repositories without also checking out data from Git LFS.


SHyx0rmZ:
Fixes #79.

philk:
Hah, I have a branch that's identical to this that I've been testing locally except I named the parameter `disable_git_lfs`. 👍

SHyx0rmZ:
Hm, inverting the logic didn't occur to me when I added this to fix our pipeline. I think I like `disable_git_lfs` more, because it should also simplify the `jq` filter. Will change it tomorrow.

vito:
Change looks good, but I'd like to understand why you need this before adding another knob. :) Is it similar to the reason for skipping submodules? (i.e. optimization)

philk:
We have around 2GB of files in git-lfs that we don't need for tests, just dev so it's to avoid pulling those down in CI.

SHyx0rmZ:
Similiar situation here. We have a pipeline that tests all commits to the configuration management repository. Not fetching from Git LFS (unless required) saves us lots of traffic and (more importantly) build time.

vito:
Sounds reasonable. Thanks!

PR #86, opened 2016-10-13
Allow overriding of git branch when pushing to a repository

factorypreset:
Sometimes we need to be able to push code to a new branch that we determine in the course of a pipeline. This change allows us to override the source configuration's branch name when pushing to the repository.

To test this:
 * define a git-resource with a given branch, and set the new `all_branches: true` source option
 * in a task, write out a file containing the name of a new branch (e.g. some random string)
 * do a `put` to the git resource, setting the `branch: path/to/the/file` param

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,9 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
 * `gpg_keyserver`: *Optional*. GPG keyserver to download the public keys from.
   Defaults to `hkp:///keys.gnupg.net/`.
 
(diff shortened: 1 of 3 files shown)


vito:
Two problems I see with this conceptually:

* `all_branches: true` introduces unversioned data to the fetched bits, making the caching semantics impossible to define cleanly; you'll have old branches lying around. You should have a different resource for the other branch instead.

* For dynamic branch specification, I'd rather tackle https://github.com/concourse/concourse/issues/684 and resist adding more and more ways to use our various resources.

Thanks though!

PR #90, opened 2016-10-27
Show a commit url for github repo

karampok:
closes concourse/concourse/issues/430

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -99,12 +99,28 @@ add_git_metadata_message() {
   ]"
 }
 
+add_git_metadata_url() {
+  local commit=$(git rev-parse HEAD)
+  local org=$(git remote get-url origin|grep github| awk -F'[/:]' '{print $(NF-1) }')
+  local repo=$(basename `git rev-parse --show-toplevel`)
+  local url="https://github.com/$org/$repo/commit/$commit"
+
+  if [ -n "${org}" ]; then
+    jq ". + [
+      {name: \"commit_url\", value: \"${url}\"}
+    ]"
+  else
+    cat
+  fi
+ }
+
 git_metadata() {
   jq -n "[]" | \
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@karampok Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=90) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

databus23:
Would be rad if this would also work for github enterprise installations. Our GHE instance host a hostname of github.organiztion.corp. Right this looks like it would create a link for github.com which would yield a 404.
I think the hostname shouldn't hardcoded but extracted from the remote url.

vito:
Agreed with @databus23 - we've got a lot of people using GHE, so it'd be nice to have it work for them too.

gabro:
@databus23 @vito we use GHE support too, and I agree with you. Extracting the GHE URL could be something like

```bash
local hostname=$(git remote get-url origin | awk -F'[/:]' '{print $(NF-2) }' | awk -F'@' '{print $(NF)}')
```

Unfortunately there's no obvious way to make sure this is an actual GHE instance.

The `github.com` detection works by grepping on the `github` string, but for GHE there's no such marker, so this would to produce 404 URLs for every repo not on GitHub (nor .com nor enterprise).

I think the only way to get this done reliable would be to add an extra `isGithub` flag to the resource.

Thoughts?

chendrix:
This has been open for 7 months and the CLA has not been signed and the issues around GHE have not been addressed, so I'm going to close this.

If @gabro you are interested in copying this PR and getting it to a mergeable state, I'd be more than happy to look at it again.

Thanks all for your inputs!

PR #97, opened 2016-12-01
Add the ability to force push a branch.

dfedde-pivotal:
Add the 'force' option to the put script allowing the user to force
the branch to be pushed to the remote.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -33,6 +33,7 @@ repository=$(jq -r '.params.repository // ""' < $payload)
 tag=$(jq -r '.params.tag // ""' < $payload)
 tag_prefix=$(jq -r '.params.tag_prefix // ""' < $payload)
 rebase=$(jq -r '.params.rebase // false' < $payload)
+force=$(jq -r '.params.force // false' < $payload)
 only_tag=$(jq -r '.params.only_tag // false' < $payload)
 annotation_file=$(jq -r '.params.annotate // ""' < $payload)
 
@@ -60,6 +61,11 @@ if [ -n "$tag" ] && [ ! -f "$tag" ]; then
   exit 1
 fi
 
+forceflag=""
+if [ $force = "true" ]; then
+  forceflag="--force"
+fi
+
 tag_name=""
 if [ -n "$tag" ]; then
   tag_name="$(cat $tag)"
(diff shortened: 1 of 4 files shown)


test/put.sh
```diff
@@ -293,6 +293,33 @@ it_can_put_and_set_git_config() {
...
+  cp ~/.gitconfig ~/.gitconfig.orig
+
+  put_uri_with_force $repo1 $src repo | jq -e "
+    .version == {ref: $(echo $ref | jq -R .)}
+  "
+
+  # switch back to master
+  git -C $repo1 checkout master
+
+  test "$(git config --global core.pager)" == 'true'
```
vito: don't need these here - some of the rest might also be copypasta. maybe just base it on the simplest push case?

assets/out
```diff
@@ -79,7 +80,11 @@ tag() {
...
 push_src_and_tags() {
-  git push --tags push-target HEAD:refs/heads/$branch
+  if [ "$force" = "true" ]; then
+    git push --tags push-target HEAD:refs/heads/$branch --force
+  else
+    git push --tags push-target HEAD:refs/heads/$branch
+  fi
 }
 
 push_tags() {
```
vito: may also be worth updating this for the tag-only push case.

maybe just extract a `$forceflag` var that's either ` --force` or empty, and interpolate it into the arg string for both `push`es?

test/put.sh
```diff
@@ -293,6 +293,33 @@ it_can_put_and_set_git_config() {
...
+  local repo1=$(init_repo)
+
+  local src=$(mktemp -d $TMPDIR/put-src.XXXXXX)
+  local repo2=$src/repo
+  git clone $repo1 $repo2
+
+  local ref=$(make_commit $repo2)
+  make_commit $repo1
+  # cannot push to repo while it's checked out to a branch
+  git -C $repo1 checkout refs/heads/master
```
vito: imo it'd be clearer if this tested the more common case of force-pushing over commits - this is kind of a lesser-known behavior of local directory pushes

assets/out
dfedde-pivotal: Re: updating the `tag-only` push case -- would you ever need to force push if you're only updating tags?

test/put.sh
dfedde-pivotal: We thought that is what is happening. We are making a commit on both sides and should lose the commit in repo1. Would checking that the commit made in repo1 is gone in repo2 satisfy this concern?

assets/out
vito: Yeah - you need to force-push if the tag already exists.

test/put.sh
vito: Oh my bad - I misread and thought this commented line was setting up the force-push case. I see now that it's doing exactly what I described.

dfedde-pivotal:
We tried to address all of the issues you had. would you take a look at it again?

vito:
looks great, thanks a bunch!

PR #98, opened 2016-12-06
Strip newline from private key output

jmcarp:
Treat an empty private key (`""`) the same way as an undefined private key. The use case is to allow users to specify an empty private key without requiring them to drop the parameter from the resource entirely.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -3,7 +3,7 @@ export TMPDIR=${TMPDIR:-/tmp}
 load_pubkey() {
   local private_key_path=$TMPDIR/git-resource-private-key
 
-  (jq -r '.source.private_key // empty' < $1) > $private_key_path
+  (jq -r -j '.source.private_key // empty' < $1) > $private_key_path
 
   if [ -s $private_key_path ]; then
     chmod 0600 $private_key_path


pivotal-issuemaster:
@jmcarp Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=98) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

vito:
I suspect this would break the normal case of having a private key specified. In my experience the trailing linebreak is important.

pivotal-issuemaster:
@jmcarp Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=98)!

chendrix:
@jmcarp have we done the manual testing to verify this doesn't break all use cases of the private key?

jtarchie:
@jmcarp, who has an empty private key that they need to specify it?

If this a templating issue, where you can't user conditional logic because of `--load-vars`, I'd switch to using ERB to add the logic. My worry is that this could introduce "we need to do it for other config", too.

vito:
I think you could do '-y foo=nil' with the new params.

jtarchie:
We are closing this because the new templating should give you more control of YAML values that can be used.

Thanks!

PR #119, opened 2017-05-14
Add a param to not fetch recursively submodules

ArthurHlt:
Add a `disable_recursive_submodules` parameter, if `true` the `in` resourse will not fetch submodules recursively.

We sometimes doesn't need to fetch all submodules, if we don't fetch them we can gain times during task.

Usecase:
We have a bosh release like diego-release which has script and manifests to generate a final manifest, we only need those files and set as submodules this release.
But when using this resource we will fetch all its submodules which is useless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,8 @@ Submodules are initialized and updated recursively.
 
 * `disable_git_lfs`: *Optional.* If `true`, will not fetch Git LFS files.
 
+* `disable_recursive_submodules`: *Optional.* If `true`, will not fetch submodules recursively.
+
 #### GPG signature verification
 
 If `commit_verification_keys` or `commit_verification_key_ids` is specified in
--- a/assets/in
+++ b/assets/in
@@ -37,6 +37,7 @@ commit_verification_key_ids=$(jq -r '(.source.commit_verification_key_ids // [])
 commit_verification_keys=$(jq -r '(.source.commit_verification_keys // [])[]' < $payload)
 gpg_keyserver=$(jq -r '.source.gpg_keyserver // "hkp://keys.gnupg.net/"' < $payload)
 disable_git_lfs=$(jq -r '(.params.disable_git_lfs // false)' < $payload)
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@ArthurHlt Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=119)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=119) to manually synchronize the status of this Pull Request.

ArthurHlt:
Is that a new tool ? Orange already signed the CLA and it's first time so far that I see this message.

chendrix:
@ArthurHlt nope, we've always had a CLA bot.

ArthurHlt:
@chendrix Thank your for your answer but I meant, I that a new CLA bot? Orange already signed the CLA in 2016, do we need to sign an updated CLA or something ?

chendrix:
Hi there,

After reviewing your PR it looks like your use case is already covered with the `submodules` parameter. Here's the documentation for it:

`submodules`: Optional. If none, submodules will not be fetched. **If specified as a list of paths, only the given paths will be fetched.** If not specified, or if all is explicitly specified, all submodules are fetched.

(emphasis mine)

I think I'm going to close this PR unless there's a compelling use case that isn't captured with the existing parameter.

Thank you again for your contribution!

ArthurHlt:
Well it doesn't cover this usecase it's when a submodule has submodules. I've targeted one submodule specifically but this submodule has also submodules but i don't want fetch them.

chendrix:
I don't think we would want to add this particular piece of configuration as it's a very unique optimization.

poblin-orange:
@chendrix  Hello. Well it seems to be a quite common situation. My use case is including cf-release github repo in my own manifest repo, to get the manifest script.
Own Repo => cf-release-repo.
The current concourse release forces me to fetch all cf-release submodules, takes 10 min, just to get 2 shells scripts.
Same use case when including cf-redis as submodule and so on ...

Issue #120, opened 2017-05-30
paths parameter does not correctly deal with ** glob patterns


philwebb:
When trying to use the following `paths` pattern the `check` script sometimes returns the wrong commit:

```yml
      paths:
        - "**/pom.xml
```

I think this has something to do with the way that glob patterns are being expanded in the script. You can replicate the issue by jumping into the docker image and running the following

```
git clone https://github.com/philwebb/spring-boot
cd spring-boot
git checkout concourse-glob-issue
```

To simulate the existing script:
```
paths_search="-- **/pom.xml" && git log --name-only $paths_search
```

You should see `07fb9f233b71b99587c4a5152d392edb39b7b4b6`

To get the actual expected commit:
```
paths_search="**/pom.xml" && git log --name-only -- "$paths_search"
```

You should see `b55868650b1ed199f4fb64fb048f83799b3ad678`

My best guess is you need to change [this line](https://github.com/concourse/git-resource/blob/096e9228292cf0342a8b748b8cf34ecc3edb6cc0/assets/check#L81) to end `-- "$paths_search"` and then remove `--` from [here](https://github.com/concourse/git-resource/blob/096e9228292cf0342a8b748b8cf34ecc3edb6cc0/assets/check#L63)

ThibaultDelaune-pro:
I have a similar issue with glob patterns.

I try to exclude every markdown file with this:
```
ignore_paths:
- "**/*.md"
```
... but it doesn't work (jobs are still triggering when I update my readme files)!

Any idea/suggestion?

_Concourse v.7.4.0_

PR #122, opened 2017-06-01
Disable glob expansion when checking paths

philwebb:
Update the `/check` script to disable bash glob expansion of path
patterns.

As of Git 1.8.4 the platform fnmatch(3) function is used for pathspec
matching allowing patterns like "**/file" to match both "a/b/file" and
"a/foo". Unfortunately it's also possible for bash to expand these
patterns which can lead to unexpected results.

For example, prior to this commit, given files added to git in the
following order:

- 'a/file'
- 'a/b/file'

Calling the `/check` script with the pattern `**/file` would return
the commit commit containing `a/file` instead of the expected later
`/a/b/file` commit.

Fixes gh-120

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -78,6 +78,6 @@ if [ -n "$tag_filter" ]; then
   } | jq -R '.' | jq -s "map({ref: .})" >&3
 else
   {
-    git log $ci_skip --format='%H' --first-parent $log_range $paths_search
+    set -f; git log $ci_skip --format='%H' --first-parent $log_range $paths_search; set +f
   } | jq -R '.' | jq -s "map({ref: .})" >&3
 fi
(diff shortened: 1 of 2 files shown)


pivotal-topher-bullock:
Thanks for the PR @philwebb 

Hmm, I'd prefer to see this to be solved by explicitly single-quoting the paths we send to the git command, rather than toggling shell flags for file globbing around the git command. 

Totally open to suggestions as to why it should be otherwise.

/cc @vito

philwebb:
@pivotal-topher-bullock I had a brief attempt doing it that way but I didn't get very far. The problem I had was that the `paths_search` variable contained `--` and a bunch of exclusions. I couldn't work out how to quote them correctly.

My Bash skills aren't that great so there probably is a better way. The test should hopefully be useful regardless of whatever the final fix is.

pivotal-topher-bullock:
@philwebb Cherry-picked this over to `master` with the `set -f` and backfilled some other tests. 

Thanks for the PR!

PR #123, opened 2017-06-06
Fix git tag behaviour

Wintermute1:
Git resource tags were not recognized when using the tag_filter parameter. My use case required a build to be triggered on each new tag creation, for example v1.7. 

The regex v* did not trigger on the latest version.

This minor change fixed the problem.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -70,11 +70,9 @@ fi
 
 if [ -n "$tag_filter" ]; then
   {
-    if [ -n "$ref" ]; then
-      git tag --list "$tag_filter" --sort=creatordate --contains $ref
-    else
-      git tag --list "$tag_filter" --sort=creatordate | tail -1
-    fi
+
+    git tag --list "$tag_filter" --sort=creatordate | tail -1
+
   } | jq -R '.' | jq -s "map({ref: .})" >&3
 else
   {
--- a/assets/check
+++ b/assets/check
@@ -34,7 +34,7 @@ destination=$TMPDIR/git-resource-repo-cache
 
 if [ -d $destination ]; then
   cd $destination
-  git fetch
+  git fetch --tags
   git reset --hard FETCH_HEAD
 else
   branchflag=""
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@Wintermute1 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=123)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=123) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Wintermute1 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=123)!

pivotal-topher-bullock:
@Wintermute1 thanks for the PR! Is it possible to write a test for this?

DoingItNow:
It would be great if the tests can be made so this change can be pushed in as soon as. It is a change that can fix how tagging works for git-resource

Wintermute1:
@DoingItNow  I will add the tests over the weekend.

DoingItNow:
@Wintermute1 No rush! Just that nothing was happening for 3 weeks so I was wondering if the PR was dead

Wintermute1:
any feedback on the test added? Is this sufficient?

assets/check
```diff
@@ -70,16 +70,12 @@ fi
 
 if [ -n "$tag_filter" ]; then
   {
-    if [ -n "$ref" ]; then
-      git tag --list "$tag_filter" --sort=creatordate --contains $ref
-    else
-      git tag --list "$tag_filter" --sort=creatordate | tail -1
-    fi
+    git tag --list "$tag_filter" --sort=creatordate | tail -1
```
chendrix: I'm worried about this change that gets rid of the looking for `$ref`. Can someone look at the implications of this?

Wintermute1: @topherbullock @DoingItNow any suggestions?

jtarchie: I might see the error you were having.

The tags that get detected are only tags that *purposely* have commits off of a specific branch. If you were missing a tag that means it was a SHA that did not exist in the `branch` configured. This is to ensure that consistent versioning is maintained with concourse and SHAs it knows about.

This PR looks for all tags, which would include tags that weren't made against consistent branch.

@Wintermute1 can you confirm the tag that was missing was not made on a SHA that exists in your configured `branch` or the default `master`?

jtarchie:
On further discussion to my points, this would break other pipelines relying on the tags only for a specific branch. We are going to close this and update the documentation to make clear this is the intended bahaviour.

Wintermute1:
Ok thanks.

Issue #115, opened 2017-03-31
Enable ssh key verification for hosts


pronoiac:
_This is a proactive hardening measure; this is not a reaction to a current issue._

The resource currently doesn't verify the ssh key when it performs git clones - [reference.](https://github.com/concourse/git-resource/blob/d86bd0a8e7da75650ee4e0a08d6af28746754ffb/assets/common.sh#L18) I'd like to be able to enable ssh key verification. The optional parameters I see being useful: 
* enable ssh key verification
* acceptable keys, whether in [fingerprint form](https://help.github.com/articles/github-s-ssh-key-fingerprints/) or the form found in `~/.ssh/known_hosts` or the output of `ssh-keyscan`

I'd be open to hardcoding the Github keys, though I understand if that seems too much of a special case to you.

gcmalloc:
Are we trying to keep retro compatibility with that feature ? Do we want to ignore ssh fingerprint checking unless it was explicitly defined ?

Issue #84, opened 2016-10-07
Create a tunnel for using ssh behind proxy git connections


next-carlosleon:
Hi,

In our company is mandatory to use a proxy to connect to internet sites. We have a git repository in internet and only configured for ssh access, so we need to create a tunnel using corkscrew across the proxy (http://www.techrepublic.com/blog/linux-and-open-source/using-corkscrew-to-tunnel-ssh-over-http/).

We tested in ours concourse workers instances and it is working, but for using that in git-resource we need to setup the ssh config.

Before to create a fork to implement this functionality and use it, we want to know if we could implement and donate this code to this repository, and which are the steps to follow to colaborate.

We think that this functionality is useful to big companies with a complex environments, so if you want we can colaborate.

What do you think? Could you answer the questions above?

Thanks,
Carlos León

vito:
I'm honestly not familiar enough with this use case to know whether it'd be a good idea or common enough case to have in the core. Best bet may be to just fork and submit a PR and if it's not too crazy maybe we'll pull it in. Either way it sounds like you'll need to fork it.

kevinzurek:
+1 for this. Assuming PR#121 was accepted how would you use a tool like netcat or corkscrew since they aren't in the git-resource docker container?

jtarchie:
I'm pleading the @vito on this one. Providing a configuration that is the wildcard configuration, which takes any config without structure, makes it hard to change in the future. @vito articulates it well on this [comment](https://github.com/concourse/cf-resource/pull/31#issuecomment-321286995).

@kevinzurek are you willing to define a more explicit interface for the ssh config options that you need?

jtarchie:
@kevinzurek, can someone explain what corkscrew is? We honestly have no idea what some of these things are.

jghiloni:
@jtarchie I can chime in -- `corkscrew` is a tool for allowing you to tunnel SSH over an HTTPS proxy. We're running into the same problem at my client and I've verified that corkscrew works for our situation. It's a relatively small C binary that is available in most mainstream Linux package managers and seems easy to build for others (like busybox). I'd love to see this in the image, so count me down as a big +1 on this.

jghiloni:
I did a quick investigation this morning on getting the binary into the resource (as it's not in the alpine package manager). Would be willing to submit a PR. I could envision an extra block in the `source` like this:

```
source:
...
  ssh_proxy:
    host:
    port: 
    username: (optional)
    password: (optional)
```

PR #137, opened 2017-08-11
Add support for ssh config

piersy:
Additional to specifying an ssh private key users can now specify ssh
config as found in '\~/.ssh/config' and a set of known hosts as specified
in '\~/.ssh/known_hosts'.

There is a test included in this pull request that validates that the git resource can retrieve a git submodule over a proxied ssh connection. This is the use case that I have.

Fixes
- #115
- #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,29 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
     ```
     Note: You can also use pipeline templating to hide this private key in source control. (For more information: https://concourse.ci/fly-set-pipeline.html)
 
+* `ssh_config`: *Optional.* SSH config to use when pulling/pushing. If this
+   if this is not specified a default ssh config that disables strict host 
+   key checking will be used. 
+    Example:
+    ```
+    ssh_config: |
+      Host 10.34.100.110
+      	StrictHostKeyChecking no
+      
+      Host gitserver
+    	ProxyCommand ssh -q proxyuser@94.14.162.172 nc 183.66.155.40 22
+    ```
+    Note: You can also use pipeline templating to hide this config in source control. (For more information: https://concourse.ci/fly-set-pipeline.html)
+
+* `known_hosts`: *Optional.* SSH known_hosts file to use when pulling/pushing.
+    Example:
+    ```
+    known_hosts: |
+      |1|KCXSa0KJnCIzB5UucUO... <more random chars> ...9EZvpaqtbxhZk8DR+xrI=
+      |1|01LwurJMVF6YVwfCv9o... <more random chars> ...ASnaJocGBNYwieuBC49U=
+    ```
+    Note: You can also use pipeline templating to hide the known hosts in source control. (For more information: https://concourse.ci/fly-set-pipeline.html)
+
 * `username`: *Optional.* Username for HTTP(S) auth when pulling/pushing.
   This is needed when only HTTP/HTTPS protocol for git is available (which does not support private key auth)
   and auth is required.
--- a/assets/check
+++ b/assets/check
@@ -15,7 +15,7 @@ payload=$TMPDIR/git-resource-request
 
 cat > $payload <&0
 
-load_pubkey $payload
+configure_ssh $payload
 configure_git_ssl_verification $payload
 configure_credentials $payload
 
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -1,24 +1,29 @@
 export TMPDIR=${TMPDIR:-/tmp}
 
-load_pubkey() {
-  local private_key_path=$TMPDIR/git-resource-private-key
+# Configures ssh keys and config are placed in the default location under ~/.ssh
+# rather than under /tmp. This makes writing the ssh cofig much easier.
+configure_ssh() {
+  rm -rf ~/.ssh
+  mkdir -p ~/.ssh
 
-  (jq -r '.source.private_key // empty' < $1) > $private_key_path
+  (jq -r '.source.private_key // empty' < $1) > ~/.ssh/id_rsa
+  (jq -r '.source.ssh_config // empty' < $1) > ~/.ssh/config
+  (jq -r '.source.known_hosts // empty' < $1) > ~/.ssh/known_hosts
 
-  if [ -s $private_key_path ]; then
-    chmod 0600 $private_key_path
+  if [ -s ~/.ssh/id_rsa ]; then
+    chmod 0600 ~/.ssh/id_rsa
 
     eval $(ssh-agent) >/dev/null 2>&1
     trap "kill $SSH_AGENT_PID" 0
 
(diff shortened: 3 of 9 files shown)


pivotal-issuemaster:
@piersy Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=137)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=137) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@piersy Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=137)!

jtarchie:
I think this is valuable feature. I'd like to have the configuration be more explicit rather than *everything but the kitchen sink* configuration. @vito has a [great explanation](https://github.com/concourse/cf-resource/pull/31#issuecomment-321286995) of the reasoning.

In favor of explicit config, what exactly do you need from here. `known_hosts` seems like a format that can be extracted out. `ssh_config` can provide lots of things can we tighten that up?

andrew-edgar:
Interesting I was just looking at something similar except our use case is we have several submodules which all need different private keys.

I was thinking of creating a PR so that you can load multiple private keys instead of only 1.  Would this be of interest either as part of this PR or a separate PR?

piersy:
Hi @jtarchie I think this case is different from the situation that @vito  was commenting on.

In that case a catch all parameter was being used to specify a selection of arguments to docker, having a collection of program arguments grouped together is what caused the issue when the resource was refactored. The refactoring highlighted the need to be able to address each argument individually and understand which argument it was. In this case I believe the approach used is actually in line with @vito's advice.

> Resources should have high-level settings in params for everything, and understand every input. 

This is a high level setting in params that is essentially the `-F` parameter of ssh. 

This resource doesn't need to interpret that parameter any more deeply than knowing it is the ssh config because ssh will parse it. Git will not change their reliance on ssh because that would break git for a huge amount of people. So as long as you use git you can be sure that ssh config will be supported.

The only alternatives I see to using git is to use libgit2 or write your own git client from scratch. Neither of these are appealing options since both will always be behind the main line of git development and constantly be playing catch-up. In any case there is nothing wrong with using the git binary as currently happens in this resource.

I also think that allowing users to be able to simply drop in a bit of config that they have pre-existing is a much nicer mechanism than requiring them to break down their existing config and have to re-assemble it in a different structure defined by the git resource.

I do however think that currently this is not quite complete, I'm in agreement with @andrew-edgar that we should add support for multiple named ssh keys.

jtarchie:
@piersy,thanks for clarifying, but it is not a different case.

`ssh_config` has a lot of features that may not be supported on the docker image for the resource. In your example, you define the `ProxyCommand` this can be setup many different ways using `ssh`, `socat`, [corkscrew](https://github.com/concourse/git-resource/issues/84#issuecomment-321374794), etc. This PR doesn't provide any validation, error messaging, or guarantees for our support contract for this configuration value.

If we explicitly defined ssh proxying then we can write tests that we don't break that feature.

piersy:
Hi @jtarchie 

> This PR doesn't provide any validation, error messaging, or guarantees for our support contract for this configuration value.

Could you elaborate on this? What is a support contract here in this instance, what kind of validation or error messaging are you referring to?

jtarchie:
@piersy, my point is that there is none. Since we'd allow a kitchen sink configuration, we don't know if we'd break something for someone. The `ProxyCommand` was just an example, it wouldn't provide any error messaging that it work or didn't work.

piersy:
@jtarchie Can you give an example of a support contract that you have defined for some other resource parameter?

vito:
Closing this as it's an example of the [`kitchen-sink-config`](https://github.com/concourse/concourse/wiki/Anti-Patterns#kitchen-sink-config) anti-pattern.

@piersy All we're asking is to take one step back from the PR as-is, to what the original goal was, and add an interface that supports it at a higher level that's easier support. This could be done by adding an explicit `known_hosts` param.

Thanks for submitting this (and double thanks for adding tests), we just don't want to expose such a broad vector of configuration, as it makes it harder to safely make changes to the resource. For example, if we switch `check` to a pure-Go version that doesn't shell out to `git`, anything configured with a wildcard `gitconfig` setting no longer applies. But if everything is explicit, we can know to continue supporting `known_hosts` (or whatever other params are present`).

piersy:
Hi @vito.

I can understand that the more configuration options you add the more work needs to be done when switching implementations.

I don't see that breaking these down into individual params helps though. For example if I were to submit a pull with an explicit param for every feature available in an ssh config file, the work to migrate the implementations would be the same as if there was one param that specified an ssh config file. 

The work required is the work to support each option whether it was specified explicitly or if it was part of a config file. In the case of explicit params you will need to check all your params to see what must be done, in the case of a config file you must check the specification for the specific version of ssh that you are using.

It seems to me that this project already has "kitchen sink" configuration since you have the 

* `git_config`: *Optional.* If specified as (list of pairs `name` and `value`)
  it will configure git global options, setting each name with each value.

Its a list of name value pairs, but the names and values are not explicitly defined by the project. They are implicitly based on the version of git that is used in this resource.

When you switch your git implementation you will need to check the spec for the version of git in use in the resource to  determine what git options must be supported. The fact that they are specified as a name value pair in the yaml doesn't reduce the work that must be done (apart from writing a parser).

If the issue here is that writing parsers is too much work, or that clustering options inside config files is incongruous when switching to an implementation that does not use those config files then I understand. 

But I'm still struggling to follow this idea of explicitness, given your use of git_config inside this resource.

vito:
@piersy

> I can understand that the more configuration options you add the more work needs to be done when switching implementations.

This isn't the point. The pain is in adding *opaque* configuration objects that are 1. so tightly coupled to the tooling and 2. so powerful that users will reach for it before asking for a proper feature narrative. Supporting 10 params is just about as easy as supporting 5. That's why we prefer to add them.

> However I don't see that breaking these down into individual params helps. For example if I were to submit a pull with an explicit param for every feature available in an ssh config file, the work to migrate the implementations would be the same as if there was one param that specified an ssh config file.

Your PR is not to add every single config option. You are adding two, for your specific use cases. If someone were to submit a PR to explicitly add params for every single SSH config, I would then have them validate their use case for each and every one, because I certainly don't want to support every one of them, for the same reason I don't want to merge a PR that supports setting SSH config directly.

With each feature explicitly justified by a user need, supported and documented, we can know to not break them, or deprecate them with a warning, etc. etc., and be more flexible in our interpretation of them since we understand the format (rather than having to use a SSH config parser and reference their man pages).

> [...] But I'm still struggling to follow this idea of explicitness, given your use of git_config inside this resource.

You're completely right! `git_config` is a perfect example of something we should *not* have merged. In pointing out the problems with it you've outlined the exact reasons we don't want to merge in PRs like that ever again.

The `git_config` option came in before we were bitten by this anti-pattern and had the chance to process it.

PR #138, opened 2017-08-12
Request-for-comments: Add git tag signature verification

keymon:
What?
----

We want to implement the case of adding verification the tags instead of commit only. 

Our specific use case is that we want a human to sign a tag to give the "seal of approval" of some code to be automatically deployed into prod.

My intention is keep the current behaviour and be fully backwards compatible, but optionally disable the commit verification and enable tag verification. When verifying the tags, I think it should be enough to verify that at least one tag is signed, as usually a commit can have multiple unsigned tags and one signed tag.

I think the setting should be passed as a parameter in the `get:`. That would allow the user to use different verifications in different jobs, or even disable the verification for some cases.

In this PR I propose one implementation, but there are different alternatives and I would like to request feedback first.

Option 1
--------

(Currently implemented)

Allow pass the parameter `verify` with a list of elements to verify. Valid values are "commit" and "tags".

"commit": Will verify the current commit signed, meanwhile "tags" will verify that AT LEAST one tag is signed with the provided keys.

The default value is `verify: ["commit"]`, which keeps the current behaviour. If the user passes `verify: []` it would disable any the verification.

Option 1.b: `verify` can get "commit", "all_tags" or "any_tag"

Option 2
--------

Allow pass two new attributes:
 - verify_commit: true|false. Default: true.
 - verify_tags: true|false. Default: false

Option 2.b: do also `verify_all_tags` `verify_any_tag`

Other comments
-------------

If we verify commits and tags, the prefix `commit_` config `commit_verification_keys` and `commit_verification_key_ids` makes less sense. Maybe we should deprecate that option and provide alternatives like:  `gpg_verification_keys` and `gpg_verification_key_ids` or similar.


pivotal-issuemaster:
@keymon Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=138)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=138) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@keymon Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=138)!

jtarchie:
I am not sure I understand this feature as described.

* What is a signed tag?
* Why is it important to have these tags?

keymon:
A signed tag in git is a tag that is signed with GPG, in the same way that a commit can be signed with GPG.

> What is a signed tag?

You can find more information [in the simple docs from github here](https://help.github.com/articles/signing-tags-using-gpg/) or in the [more explicit git docs](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work)

To sign a tag, add `-s` to your `git tag` command.
```
git tag -s mytag
# Creates a signed tag
```
Verify your signed tag it by running `git tag -v [tag-name].`
```
git tag -v mytag
# Verifies the signed tag
```

> Why is it important to have these tags?

I briefly explain a use case at the beginning of the PR, but I will elaborate.

This would allow implement different use cases where some kind of signature and verification is required. It would basically cover the same cases that signing commits, but using tags instead.

A user might be interested on use tags rather than commits because:
 - it will NOT create a new commit, preventing other integrations or pipelines from being triggered.
 - tools and automation might use tags 
 - it makes sense to use a tag rather than a empty commit when one only wants to "give an approval""

A example use case:
 - A project X uses continuous integration and deployment. Every single commit is tested, deployed and tested again in `CI` and `Staging`.
 - But in order to deploy in `Production` we want only to deploy commits explicitly approved.
 - **Using a signed tag**: A human will review the latest changes, and create a signed tag `prod-v1.2.3-signed`
 - The production pipeline will be triggered for any `prod-v*-signed`, and it will verify the commit.

# Why not using a "empty commit"?

 - Adding a new commit would unnecessary re-trigger all the pipeline and other integrations.
 - It will change the HEAD ref id used in staging vs. used in prod. 
 - concourse metadata would refer to the empty commit, breaking some automation and reporting.

keymon:
@jtarchie does my comment clarify it? Do you have any feedback about how the PR should be implemented (see PR body)?  
thx

jtarchie:
@keymon, I'm not familiar with this workflow. It makes _sense_, but cannot make a judgement call if it is **right**.  Or in emoji form `¯\_(ツ)_/¯`.

The git resource is trying to do too much. Supporting tags and branches with their workflows is polluting the git resource with lots of conditional logic.

I'm working on a [proposal](https://github.com/concourse/git-resource/issues/145) to perhaps create a git tag resource. It is a little barren as I'm tracking tag related issues.

We are talking to people about their needs and workflows for tags. Would you be willing to have a conversation with me and @jama-pivotal?

keymon:
You are right about the issues in git, and the tags make it more complicated in general. 

Also, I though about this and I think I can propose other PR that simply checks the signature on the current commit OR any tag. That would satisfy our workflow, and it would be a smaller change. 

I would have a though about what you say of a git tag resource. Regarding having a chat, sure, I would try to reach you tomorrow, but I am not sure if I will be able. And after, I would be in a trip for a pair of weeks. But I would contact you after otherwise.

PR #139, opened 2017-08-17
Populate .git/ref with the reference checked out

keymon:
When using `tag_filter`, the version reported by `check` would be the matching tag name.

But if multiple tags point to the same commit, it is not possible to know which tag in the repository is being used to checkout the code in concourse.

By writing the value of `$ref` in `.git/ref` the consumers of the resource can know the exact version used by concourse, regarless it is a tag or a commit SHA-1.

Example use case
----------------

We add additional metadata about the code using tags as it passes through our pipeline. These tags trigger other pipelines.

But we cannot distinguish what tag is being checkout by concourse if multiple tags point to the same commit.

Currently we can workaround the issue by using different heuristics (date, name regex) or querying the concourse API. It would be far easier with a file container the $ref.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,11 +151,16 @@ the source configuration, it will additionally verify that the resulting commit
 has been GPG signed by one of the specified keys. It will error if this is not
 the case.
 
-#### Committer notification on failed builds
+#### Additional files populated
 
-For ease of use there is a special file `.git/committer` which is populated with
-the email address of the author of the last commit. This can be used together with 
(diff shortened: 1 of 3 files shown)


jtarchie:
Thanks for the PR! We accepted your other PR and now there is a conflict. Let me see if I can fix it.

jtarchie:
Rebased and merged (77defe11dc19dcefa4d36f6c85974c685f5cc1f3). Thanks for the PR!

PR #140, opened 2017-08-17
Fix test for committer_email

keymon:
The test will never fail because the error return code of the `test` is
being masqueraded by the echo command.

We use a subshell that would `return 1` to break the execution and make
te test function return an error code.


vito:
thanks!

PR #154, opened 2017-10-28
Add support for SSH-based access to git via HTTPS Proxy

jghiloni:
This addresses #84 based on https://github.com/concourse/git-resource/issues/84#issuecomment-339969265 and https://github.com/concourse/git-resource/issues/84#issuecomment-339989260 

The original design discussed was to use `corkscrew` but it is very sparsely maintained and gave both me and @jtarchie a bad vibe. After some discussions with him, we decided to use [ProxyTunnel](https://github.com/proxytunnel/proxytunnel) instead. This fix adds it to the base image, as well as adding the necessary configuration to make it work in `check`, `in`, and `out`. To be more explicit, we also renamed the `ssh_proxy` block described in the comment to `https_tunnel`, since that's really what this is.

Because this type of work doesn't lend itself to unit testing, I added some very simple ones to simply ensure that `proxytunnel` was successfully installed. In addition, I added a suite of basic integration tests that can be run if a repo and private key are available that stands up a `squid` proxy and attempts to do basic IO to the specified repo. If they are not specified, the integration tests are skipped.

I created a standalone github user and repo to do this testing (as my original plan was to include those credentials with the tests, but the more I thought on it, the more it sounded like a really terrible idea and I made the integration tests optional instead). I would be happy to share those credentials with the team if they want a quick repo to do validate this testing themselves.


vito:
This is pretty much an exemplary PR. Thanks for all the info - it really helps. I'll go ahead and merge, but I do wonder if we could make the integration suite run against an in-container SSH server/git daemon or something.

PR #160, opened 2017-11-27
Make proxy_host and proxy_port (sort of) optional

jghiloni:
If you are trying to run a pipeline on two different Concourse installations, and SSH proxy is required for one but not the other, the `https_tunnel` source config will fail, as it will generate a bad `ProxyCommand` statement in `~/.ssh/config`. This fix (with accompanying test) simply ensures that both `proxy_host` and `proxy_port` are set, or the entire `https_tunnel` block becomes a no-op


vito:
thanks!

Issue #158, opened 2017-11-16
fatal error when shallow cloning repo with a path filter


bhcleek:
Given a git resource with a path filter, a repository that has more recent commits that do not satisfy the filter, and a `get` that limits the depth (`1` is the ideal value to try), the `get` will fail with a fatal error, because the `.version.ref` is not present in the shallow clone.

The problem is here: https://github.com/concourse/git-resource/blob/master/assets/in#L61-L66.

I don't have time to submit a well-tested patch now, but changing those lines to 
```
git clone --single-branch $depthflag $uri $branchflag $destination

cd $destination

git fetch origin refs/notes/*:refs/notes/*
git fetch $depthflag $ref
git checkout -q $ref
```

may well fix the problem.

bhcleek:
In addition to the issue of a git resource filter on the path, I realized that this same behavior can manifest itself due to a race condition between the time a new version is detected and the time the job runs; in the interim, another commit could have been pushed, causing the same fatal error.

timrchavez:
@bhcleek Assuming I have the test correct, I do not think your change will work. I'm getting ```running it_can_shallow_clone_ref_with_new_commits...
  Switched to a new branch 'bogus'
  Switched to branch 'master'
  Cloning into '/tmp/git-tests.dlHJcd/git-tests.JFhcGp/destination'...
  error: Server does not allow request for unadvertised object 57b3b83c0199e8d73a1d1921785f2f73b0eb5ddf
The command '/bin/sh -c /tests/all.sh' returned a non-zero code: 1
```

The test is adding a couple commits but setting `version.ref` to the first commit.

timrchavez:
Updated to fetch enough depth to ensure `git checkout -q $ref` succeeds.

timrchavez:
Wondering if there's any news on this being accepted (or rejected)?

timrchavez:
Totally getting back to this :)

PR #167, opened 2018-01-19
Make shallow cloning more durable

timrchavez:
If a git-resource is configured to use shallow cloning and new commits
are added to the repository that are not considered new versions of the
resource (e.g. via a path filter excludes it), a `get` will shallow
clone HEAD which will not be equal to the version ref. This will result
in breakage.

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -63,7 +63,11 @@ git clone --single-branch $depthflag $uri $branchflag $destination
 cd $destination
 
 git fetch origin refs/notes/*:refs/notes/*
-git checkout -q $ref
+until `git checkout -q $ref`; do
+    deepen=$((deepen+$depth))
+    deepenflag="--deepen $deepen"
+    git fetch $deepenflag origin
+done
 
 invalid_key() {
   echo "Invalid GPG key in: ${commit_verification_keys}"
--- a/test/get.sh
+++ b/test/get.sh
@@ -171,6 +171,20 @@ it_honors_the_depth_flag() {
   test "$(git -C $dest rev-list --all --count)" = 1
 }
 
+it_can_shallow_clone_ref_with_new_commits() {
+  local repo=$(init_repo)
+  local myCommitRef=$(make_commit $repo)
+  local newCommitRef=$(make_commit $repo)
+
+  local dest=$TMPDIR/destination
+
+  get_uri_at_depth_at_ref "file://"$repo 1 $myCommitRef $dest | jq -e "
+    .version == {ref: $(echo $myCommitRef | jq -R .)}
+  "
+
+  test "$(git -C $dest rev-parse HEAD)" = $myCommitRef
+}
+
 it_honors_the_depth_flag_for_submodules() {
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@timrchavez Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=167)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=167) to manually synchronize the status of this Pull Request.

timrchavez:
Assuming I have the test correct, I do not believe the suggested change for #158 will work. The error I'm getting is:

```
running it_can_shallow_clone_ref_with_new_commits...
  Switched to a new branch 'bogus'
  Switched to branch 'master'
  Cloning into '/tmp/git-tests.dlHJcd/git-tests.JFhcGp/destination'...
  error: Server does not allow request for unadvertised object 57b3b83c0199e8d73a1d1921785f2f73b0eb5ddf
The command '/bin/sh -c /tests/all.sh' returned a non-zero code: 1
```

bhcleek:
On a shallow clone, I am able to duplicate the error you're seeing. It appears, though, that setting `uploadpack.allowReachableSHA1InWant` on the server will allow an arbitrary SHA to be retrieved.

RE: https://git-scm.com/docs/git-fetch-pack

timrchavez:
I use a good ol `until` loop to progressively fetch deeper and deeper until I can successfully check out the ref.

assets/in
```diff
@@ -63,7 +63,11 @@ git clone --single-branch $depthflag $uri $branchflag $destination
 cd $destination
 
 git fetch origin refs/notes/*:refs/notes/*
-git checkout -q $ref
+until `git checkout -q $ref`; do
```
vito: This will result in an infinite loop if `$ref` is no longer available anywhere (i.e. someone `push -f`ed).

assets/in
```diff
@@ -63,7 +63,11 @@ git clone --single-branch $depthflag $uri $branchflag $destination
...
+    deepen=$((deepen+$depth))
```
vito: This only makes sense to attempt if `depth` is given (it defaults to 0).

assets/in
timrchavez: Hum. Good catch!

assets/in
timrchavez: :+1:

timrchavez:
@vito Yep. I should have time. I'm not sure about @bhcleek's approach honestly because it requires a configuration change to the GitHub installation itself _and_ based on the documentation it's very resource intensive. I'm open to other suggestions, but I'm not creative enough to think of anything better than a loop which fetches deeper and deeper until it can find its ref.

vito:
@timrchavez I'm fine with it fetching all the way back to the first commit in the `push -f` case and then giving up, just as long as it's an infinite loop.

vito:
merged #205

PR #92, opened 2016-10-31
Highlight secure method of storing private data

srizzling:
It is unclear to a newbie how concourse protects passwords placed in source control. I beleive we should be making using a pipeline template replace the default behavouir currrently documented (not in scope of this PR)

I know there are other better solutions like using vault (documented here: http://www.starkandwayne.com/blog/spruce-vault-concourse-you/) but I just matched the doc style (KISS)

Sriram

PR #168, opened 2018-01-20
Fix annotate file path

krishicks:
Because the annotation file's path is relative and because of the `cd $repository` that occurs on line #92, if you want to use an upstream output directory for the contents of the tag and the annotate files, you have to do something like the following:

```yaml
- put: some-git-repo
  params:
    repository: some-git-repo
    tag: tag/tag
    annotate: ../tag/annotate
```

This is surprising and probably not what was intended. You'd rather specify it like this:

```yaml
- put: some-git-repo
  params:
    repository: some-git-repo
    tag: tag/tag
    annotate: tag/annotate
```

The test for this was fixed as well. In the normal case you cannot pass the source directory as part of the path; you can only pass a path relative to the source directory.

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -69,6 +69,11 @@ if [ -n "$tag" ] && [ ! -f "$tag" ]; then
   exit 1
 fi
 
+if [ -n "$annotation_file" ] && [ ! -f $annotation_file ]; then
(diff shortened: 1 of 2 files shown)


vito:
thanks!

PR #171, opened 2018-02-03
Make Git commit message accessible

elgohr:
This pull-request makes the Git commit message accessible for tasks, so that people can share what they have done with other resources (Slack/E-Mail/Carrier pigeon).

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -137,7 +137,7 @@ echo "${return_ref}" > .git/ref
 # the content of a successfull build.
 # Using https://github.com/cloudfoundry-community/slack-notification-resource
 # for example
-echo $(git_metadata) | jq -r '.[6].value // ""' > .git/commit_message
+git log -1 --format=format:%B > .git/commit_message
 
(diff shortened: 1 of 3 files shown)


assets/in
```diff
@@ -133,6 +133,12 @@ git --no-pager log -1 --pretty=format:"%ae" > .git/committer
 # pulled ref in following tasks and resources.
 echo "${return_ref}" > .git/ref
 
+# Store commit message in .git/commit_message. Can be used to inform about
+# the content of a successfull build.
+# Using https://github.com/cloudfoundry-community/slack-notification-resource
+# for example
+echo $(git_metadata) | jq -r '.[6].value // ""' > .git/commit_message
```
vito: Rather than plucking it out of the metadata, I think it makes more sense to just get it the same way the metadata gets it (`git log -1 --format=format:%B`). That way it won't break if we add another metadata field before that one (`6`).

vito:
thanks!

PR #172, opened 2018-02-06
Update to allow dynamic branch choices during 'out'

ndmckinley:
My use case is a code generator which needs to push to a different branch every time it runs.

Changes to `check` and `in` make them work with the branch-agnostic `out`.

Tests pass, although I didn't run the integration tests.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -39,11 +39,13 @@ if [ -d $destination ]; then
   git reset --hard FETCH_HEAD
 else
   branchflag=""
+  singlebranch=""
   if [ -n "$branch" ]; then
     branchflag="--branch $branch"
+    singlebranch="--single-branch"
   fi
 
-  git clone --single-branch $uri $branchflag $destination
+  git clone $singlebranch $uri $branchflag $destination
   cd $destination
 fi
 
--- a/assets/in
+++ b/assets/in
@@ -49,16 +49,16 @@ if [ -z "$uri" ]; then
 fi
 
 branchflag=""
+singlebranch=""
 if [ -n "$branch" ]; then
   branchflag="--branch $branch"
+  singlebranch="--single-branch"
 fi
-
 depthflag=""
 if test "$depth" -gt 0 2> /dev/null; then
   depthflag="--depth $depth"
 fi
-
-git clone --single-branch $depthflag $uri $branchflag $destination
+git clone $singlebranch $depthflag $uri $branchflag $destination
 
 cd $destination
 
--- a/assets/out
+++ b/assets/out
@@ -30,6 +30,7 @@ configure_credentials $payload
 uri=$(jq -r '.source.uri // ""' < $payload)
 branch=$(jq -r '.source.branch // ""' < $payload)
 git_config_payload=$(jq -r '.source.git_config // []' < $payload)
+branchfile=$(jq -r '.params.branch_file // ""' < $payload)
 repository=$(jq -r '.params.repository // ""' < $payload)
 tag=$(jq -r '.params.tag // ""' < $payload)
 tag_prefix=$(jq -r '.params.tag_prefix // ""' < $payload)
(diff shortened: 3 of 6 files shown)


pivotal-issuemaster:
@ndmckinley Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=172)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=172) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ndmckinley Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=172)!

vito:
Thanks for submitting this!

I'm not really comfortable with the changes to `/in` and `/check` necessary to support this, however. I think this actually kind of ties together a few thematically large issues, some we're still stewing and some we have in progress:

* With https://github.com/concourse/concourse/issues/684 you wouldn't need 
a new param, since you could just dynamically populate it.
* With https://github.com/concourse/concourse/issues/534 you would be able to specify the param in the `put`, and not require it in the `source`, since resources would then just receive one big `config` object.
* With https://github.com/concourse/concourse/issues/1707 this might actually need to run against a different *space*, since that's how the git resource would represent branches. In fact, depending on what you're trying to do, spaces might be the real answer.

Could you provide an example of how this would be used? It may make more sense to focus on the aforementioned issues instead.

Thanks again!

ndmckinley:
I see why you wouldn't be comfortable with it - git has a lot of levers and knobs, working with multiple branches is hard, and this substantially broadens the state that the repo might be in after cloning.

I need to use this because I have a code generation pipeline which needs to push intermediate state from multiple pipeline runs at once without clobbering each other.  To that end I'm using a computed unique branch name which can only be known after the task generating the output has run.  It's one of the use cases specified in the "spaces" issue you posted, but my particular use case would be solved by either of concourse/concourse#684 or concourse/concourse#534.

In the meantime, since this PR won't work for you, what do you suggest for this parallel pipelines use case?

vito:
@ndmckinley Gotcha, so are you pushing the generated code to a random branch just so that you can propagate it to subsequent jobs? Could some other resource type be used, or is there some significance to using branches?

ndmckinley:
Assuming you mean random colloquially, as in "not determinable ahead of time", yep!  And unfortunately it does need to be a git repository, because a later step of the pipeline is going to open a GitHub pull request using the generated code.

ndmckinley:
Hi,

Sorry, not in this case - I see where you're going with that, and I can imagine it being useful, but my particular needs aren't met by it.

You're right that there's a good reason I'm not doing it in one giant job - we're running many acceptance tests from potentially dozens of downstream repositories, and fanning out (then back in) is the best model for the workflow.  The tests would work fine with the arbitrary artifacts you describe - in this case the artifact is the generated code and it's being streamed to test workers - it's code, so it's small, and the git repo history (which is the only large part) can be left behind with no problem.  But after all the tests, unfortunately, that artifact *does* still have to wind up on GitHub (in a branch that won't clobber any parallel runs of the pipeline), because I need to issue pull requests against the downstream (generated) repository.

Unfortunately I haven't quite understood your comment about unversioned state - my generated code is associated with a version of the codegen config repository.  Plus, if it's global state, then that's trouble for me because I'm likely to have several parallel pipeline runs at a time.  If it's per-run... as far as I know Concourse doesn't really have any notion of "per-run", but I suppose that would work well for me.

vito:
> e: Oh, I get it now! Yes, I would write a version of git-resource which was, instead of being a resource, git-state or something, and that would have the ability to push to arbitrary branches, which would solve many of the concerns you have here. Clever!

Right! This conversation has actually spurred on a lot of thinking about spaces and the new resource interface. I'll have something visible at https://github.com/concourse/rfcs some time soon.

vito:
@ndmckinley I've submitted https://github.com/concourse/rfcs/pull/1 which includes a [tentative example](https://github.com/vito/rfcs/blob/resources-v2/01-resources-v2/branch-gen.yml) for branch generation. The tl;dr is that you would `put` to a new space with a consistent prefix (i.e. branch names like `gen-*`) and then have the downstream jobs depend on those spaces via pattern matching.

If that sounds like a good direction, would you be OK with closing this and using a fork in the meantime?

Thanks!

vito:
It's a requirement of concourse/concourse#1707 which we're actively working on right now. I'd give it a few months at least, plus the time involved for migrating the resources to the new interface (which may go beyond simple migration in some cases).

Closing, thanks for all the discussion!

PR #176, opened 2018-02-28
Do a full clone when tag filter is set and branch is not

kurlzor:
I tought I was opening this against my own fork but... let's do this anyway.
Tag filtering is quite useful but being limited to the tracked branch isn't suitable for all workflows. In our team, we tag commits from other branches and expect to be able to build/deploy these tags if we want to.
Cloning with `--single-branch` prevents that. Since you can't know if the commit exists in a given branch before cloning the repository, just clone everything when there are tags involved unless you explicitly specified a branch to track.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -39,11 +39,17 @@ if [ -d $destination ]; then
   git reset --hard FETCH_HEAD
 else
   branchflag=""
+  singlebranchflag="--single-branch"
+
   if [ -n "$branch" ]; then
     branchflag="--branch $branch"
+  else
+    if [ -n "$tag_filter" ]; then
+      singlebranchflag=""
+    fi
   fi
 
-  git clone --single-branch $uri $branchflag $destination
+  git clone $singlebranchflag $uri $branchflag $destination
   cd $destination
 fi
 
--- a/assets/in
+++ b/assets/in
@@ -49,8 +49,13 @@ if [ -z "$uri" ]; then
 fi
 
 branchflag=""
+singlebranchflag="--single-branch"
 if [ -n "$branch" ]; then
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@kurlzor Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=176)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=176) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kurlzor Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=176)!

r-chris:
I was about to do a similar thing. You @kurlzor probably need to update the documentation as well. Right now the branch documentation talks about it defaulting to the `master`.

vito:
This was previously attempted in https://github.com/concourse/git-resource/pull/165 which ended up with a fairly complicated discussion. I'm not sure the proposed changes are enough.

To be honest, I don't think we should have pulled in the `tag_filter` behavior, and that should really just be a different resource. The semantics are entirely different. This may be something we consider during the migration to the new resource interface that we're planning in concourse/rfcs#1.

vito:
Hmm, I think this should maybe just be closed for now (because #165 showed that things are a bit more complicated), and we should revisit `tag_filter` in the future. It may make more sense to have support for tags across all branches as another way to support spaces (concourse/concourse#1707), either with this resource type or a more tag-focused one.

Thanks for PRing this though!

PR #181, opened 2018-03-29
Allow check to gracefully tag_filter + deleted tag

ljfranklin:
Prior behavior:
If you specified a tag_filter and the most recent tag was deleted,
`check` would start returning an empty list of commits rather than
detecting new tags

Signed-off-by: Dave Walter <dwalter@pivotal.io>


vito:
thanks!

Issue #30, opened 2016-01-28
Can't push without tags


njbennett:
A put command to a github resource is always a `git push --tag`. We use tags to mark release versions and commits deployed to specific environments. Right now, we have to delete all tags locally in our scripts before pushing from Concourse. We would prefer to be able to just specify in our plan configuration that we don't want to push tags, or for this to be the default behavior.

vito:
What's fetching with tags or creating them? We also use tags and don't experience this behavior. Agreed that it should be something you can turn off, but I'm not sure about by default; the idea is that if there are tags there it's because your script created them, so you probably want them pushed.

njbennett:
As far as I can tell, the issue is that we (Billing & Accounts Management) share our repo with another team (Apps Manager). They use tags to mark commits deployed to various CI environments and to mark commits released in various versions of PCF; they delete the tags that associate commits with CI environments in their GoCD scripts and recreate them when they push new commits to those environments.

We've always just ignored these tags, because we were pushing with GoCD scripts rather than through Concourse. Concourse is basically making us more aware of the weird thing our sister team is doing with tags, and forcing us to mirror that behavior in our CI. So this might just be something that can be chalked up to "doing something weird with Github tags + odd team/repo structure."

vito:
Gotcha. I'm just confused as to how the tags are being introduced to the local repo in the first place. When the repo is fetched, we don't fetch the tags; any tags being pushed must have been created in your script, which Concourse takes to mean that you want them pushed. Or I may be wrong, and it is fetching the tags, and normally pushing does a no-op, but if the tags have since changed remotely it'll try to clobber them?

Can you confirm that your job isn't creating any tags? I'm not totally against adding this param, I just want to make sure it's for the "right reason".

njbennett:
We're pretty sure that we're not making tags. The only times "tag" appears in our ci repo are to delete them, and in reference to our Ubuntu image.

How are you fetching without tags? It looks like git fetch, clone, and pull all bring in all the tags on the branches involved by default, and maybe even that tags can't, or shouldn't, *not* be pulled down.

vito:
Yep looks like a fresh clone does bring in the tags. I must have been thinking of something else. In that case it's probably that after cloning, the tag has changed remotely, and the `git push --tags` ends up trying to roll it back.

I would rather have the tags not be present after fetching. They're not really meant to be there, especially as the caching semantics get a little weird (it'll have cached tags that were present only when the cache is warmed).

[...]

PR #182, opened 2018-04-04
Clone repos in `in` script without tags

benmoss:
We were encountering a problem very similar to #30: when we manually deleted tags from our repo, our Concourse pipeline was then re-pushing them. We figured out that since Concourse had a cached version of our repo with the tags in it, and since it pushes with tags every time, this caused it to recreate the tags the next time the build ran.

In #30 @xoebus mentioned he'd be open to a PR that changed the resources behavior to clone without tags, but what we saw from all the tests we had to disable is that a bunch of tag metadata has been added to the output of the in script.

Hoping to start a conversation here about what to do here. Maybe @keymon should chime in here too, since he seems to have made a bunch of these tag related changes.

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -139,7 +139,10 @@ echo "${return_ref}" > .git/ref
 # for example
 git log -1 --format=format:%B > .git/commit_message
 
+metadata=$(git_metadata)
+git tag | xargs git tag -d
+
 jq -n "{
   version: {ref: $(echo $return_ref | jq -R .)},
-  metadata: $(git_metadata)
+  metadata: $metadata
 }" >&3
--- a/test/get.sh
+++ b/test/get.sh
@@ -425,6 +425,18 @@ it_decrypts_git_crypted_files() {
     ( echo "encrypted file was not decrypted"; return 1 )
 }
 
+it_does_not_retain_tags() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  git -C $repo tag v1.1-pre
+
+  local dest=$TMPDIR/destination
+
+  get_uri $repo $dest
+
+  test -z "$(git -C $dest tag)"
+}
+
 run it_can_get_from_url
 run it_can_get_from_url_at_ref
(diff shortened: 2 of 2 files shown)


benmoss:
@keymon yeah, that's reasonable, i think the problem is that we do want to sometimes push tags, but only *new* tags, as in tags that the build itself actually adds. with Concourse's caching of resources we end up re-pushing old tags if they've disappeared from the upstream.

vito:
@benmoss Yay, thanks for PRing this! This is one of those itches that we've never found time to scratch.

@keymon I think it's more correct to just never fetch them. The resource is primarily modeling *commits* as the versionable artifact, so the set of tags that are fetched will be out-of-date as they're not what's being versioned.

However I think this *might* end up breaking `tag_filter`, since the versions returned by `check` in that case will be something like `ref: <tag name>`, and that won't exist if the tags aren't fetched during the clone. The tests don't cover this because `check` and `in` are covered independently.

Here are some ideas:

* Clone the tags, but delete them all after checkout.
* When `tag_filter` is used, emit versions like `tag: <tag name>` instead and somehow check out only that tag. Maybe `git fetch origin <tag name>; git checkout FETCH_HEAD`?
* When `tag_filter` is used, emit commit shas instead of tag names. I don't really like this though because it makes it difficult to tell which commit is which tag.

benmoss:
@vito I updated my approach to just clear the tags at the end of the `in` script. LMKWYT!

keymon:
> @vito I updated my approach to just clear the tags at the end of the in script. LMKWYT!

I like this idea. One feature I would personally keep is reporting of the tags in the metadata.  We use it a lot to know which version of the repos have been pulled, and if the build executed the version we wanted. 

IIRC we had some script to parse the tags metadata of the resource to know which version was deployed (I can find it now).

This change satisfies all the requirements :+1:

vito:
Perfect, thanks! Looks like by capturing the metadata before reaping the tags, the metadata is preserved, too. Merging.

PR #191, opened 2018-05-29
Upgrade git-crypt to 0.6.0

gcapizzi:
Hi Concourse team!
We're using `git-crypt` in our pipeline and we are being hit by [this bug](https://github.com/AGWA/git-crypt/issues/71) which was fixed by [this commit](https://github.com/AGWA/git-crypt/commit/b47176e6a8b202380cb11fa3d0b2fd7331ead029). Also, `0.5.0` is very old (2015!) so it's probably worth to upgrade! Thanks 🙂 

/cc @alamages

PR #192, opened 2018-05-30
Invoke `git-crypt` directly to avoid absolute paths in .git/config

gcapizzi:
Follow on from https://github.com/concourse/git-resource/pull/191.

When `git-crypt` is invoked as `git crypt`, `git` calls `git-crypt` using an absolute path. Therefore the filter and diff config that `git-crypt` puts into `.git/config` will contain the absolute path to `git-crypt`. This is a problem for tasks that perform commits, as the `git-crypt` path in the task container may not match the one in the resource container. This allows task images to have `git-crypt` wherever they like in `$PATH`.

/cc @jamesjoshuahill

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -101,7 +101,7 @@ git submodule sync
 
 if [ -f $GIT_CRYPT_KEY_PATH ]; then
   echo "unlocking git repo"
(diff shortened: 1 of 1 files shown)


vito:
Thanks!

PR #202, opened 2018-07-26
Add ability to output short SHA with formatting

suda:
This PR allows passing `short_ref_format` parameter and creation of `.git/short_ref` file with a GitHub style short SHA. It solves a problem of building Docker images with custom tag names based on commit hash.


suda:
I see the build is failing but don't have the access to see what's wrong. Surprisingly the image does build locally, including running all the tests 🤔

suda:
I rebased this PR on top of #203 and the build passes again :)

vito:
Neat, thanks!

I am a *little* worried that we're baking more and more functionality into this resource, but this does seem really convenient and any other solution at the moment would be much too heavy. (*Maybe* in the future this kind of thing could be done with a reusable task per https://github.com/concourse/rfcs/issues/7.)

PR #203, opened 2018-07-29
libressl fixes and latest git lfs

alucillo:
1. Addresses an issue I encountered when building locally as a result of libressl
2. Upgrades git lfs to the latest version


horjulf:
Since `git` already comes from alpine, is there any issue nowadays installing `git-lfs` from alpine ?

vito:
Neat! Yeah, it looks like there's a `git-lfs` package we can use now.

suda:
I believe this PR might be the thing that fixes #202 failing to build.

alucillo:
Take a look.  I updated the PR to install git-lfs via apk now.

vito:
@alucillo Perfect, thanks!

PR #205, opened 2018-08-06
incrementally deepen shallow clones if ref not found

### LINKED ISSUE #158: fatal error when shallow cloning repo with a path filter

norbertbuchmueller:
If the given ref is not found after a shallow clone, it starts chasing it by deepening the clone in steps of 2-power, eventually falling back to a deep clone if a threshold is reached.

This should fix #158 / #167.

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -66,6 +66,46 @@ git clone --single-branch $depthflag $uri $branchflag $destination
 cd $destination
 
 git fetch origin refs/notes/*:refs/notes/*
+
+# A shallow clone may not contain the Git commit $ref:
+# 1. The depth of the shallow clone is measured backwards from the latest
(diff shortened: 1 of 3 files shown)


vito:
Awesome, thanks for tackling this! Looks good - merging.

PR #210, opened 2018-08-31
Allow specifying submodule_recursive in source config.

alext:
This adds the option for specifying `submodule_recursive` in the source
definition. This avoids the need to specify this on every get for a
given git resource in the pipeline. If this is specified in both places, the
get value will take precedence.

While testing this behaviour, I've also noticed that it's necessary to specify
this param as a string due to the way that the default value is implemented
using jq. I've added a note to the docs to clarify this.


pivotal-issuemaster:
@alext Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/git-resource&pullRequestId=210)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/git-resource&pullRequestId=210) to manually synchronize the status of this Pull Request.

vito:
Closing due to inactivity, but at least concourse/rfcs#1 should remove the need for this. Thanks again!

alext:
> Closing due to inactivity

Apologies for the delay in responding to comments here. I found that using the `or` as suggested didn't work 100% because it wouldn't allow a false at the source level to be overridden by a true on an individual get. I've finally found a way to solve this, and opened #237 with it.

PR #218, opened 2018-10-05
document the fetch parameter for the in operation

JohannesRudolph:


vito:
I think I'll just close this actually. I don't really want to draw attention to it by documenting its behavior, and now would be an awkward time to remove it, too, since we'll have a new `git` resource implementation for concourse/rfcs#1 anyway and we can just opt to not carry that feature over.

Issue #74, opened 2016-08-16
Tag behaviors should work for detached HEAD tags


nebhale:
Currently, when specifying a value for `tag_filter` the only candidates that are considered are the ones that are reachable from the specified (or default) branch.  Tags that are detached `HEAD`s (or in other branches) are not considered since they are not retrieved during `git fetch <BRANCH>`.

In Maven (and Java generally) the version of an application exists in source control.  So, when we tag a release, it requires a commit that contains those changes.  In the Spring projects, it’s common that `master` _never has a release on it_.  Version numbers go from one snapshot to another from the perspective of `master`, and the release hangs off and has to be directly targeted.

![screen shot 2016-08-15 at 15 04 34](https://cloud.githubusercontent.com/assets/60754/17717917/be3f4b6e-63c5-11e6-8ef0-5a71d7e9a72f.png)

Based on a [Slack conversation](https://concourseci.slack.com/archives/general/p1471373409005793) I'd like to advocate that any behavior having to do with tags (`tag_filter` is the only one that's obvious to me) ignore branches completely.  All tags, regardless of what they're reachable from `HEAD` or not, should be considered.  This policy should hold true for other future features having to do with tags as well.

macropin:
I think I've hit the same issue from a different angle - we're trying use git tags for docker image tags and the git resource doesn't see pushed tags unless followed by a commit. I believe this is because the git resource only runs `git fetch` not `git fetch --tags` https://github.com/concourse/git-resource/blob/master/assets/check#L37. `git fetch --tags` implies `git fetch` so a one line change should work.

ljfranklin:
Hit this as well when attempting to pull terraform's [v0.8.7 tag](https://github.com/hashicorp/terraform/tree/v0.8.7). While the `check` detected this tag, the subsequent `get` fails with:

```
Cloning into '/tmp/build/get'...
Checking out files: 100% (8552/8552), done.
error: pathspec 'v0.8.7' did not match any file(s) known to git.
```

Hijacking onto the `get` container confirms that this tag is absent, but shows up after running `git fetch --tags` as the tag exists but is not reachable from master. Related [SO answer](http://stackoverflow.com/a/15794719/1437783).

Wintermute1:
@ljfranklin i'm encountering the same issue. The check detects the tag but the get fails. Have you found a way to fix or get around this? Thanks

ljfranklin:
[...]

PR #223, opened 2018-10-26
[check] Allow triggering with tag updates

talset:
From a customer usecase, they use git tag like the docker way by updating static tag on new release:

git tag can be updating by changing the commit associeted

```
git tag DEV
git push origin DEV

# make changes
git commit

git tag DEV --force
git push origin DEV --force
```

Current git resource is only watch tags name but not tags update. So pipelines aren't triggered when the tag is updated and if the tag if from a different branch from the specified (or master) the resource fail with `error: pathspec 'DEV' did not match any file(s) known to git.`
To solve it I suggest to not use only ref but ref and commit when tag_filter is used.

This commit allow to store the tag name and commit id associeted to the tag as "version" of the resource. With thos way we are able to re-trigger the pipeline when a the tags is updated.

Here is an example or resource output with tag_filters:
```
  [
    {
      "ref": "2.0-staging",
      "commit": "73128e2cf79e5547c1ef292bc34de3b84f590b8d"
    },
    {
      "ref": "3.0-staging",
      "commit": "51af5f19c36833064c6d51d88af7495c8dcfa512"
    }
  ]
```

resource versions can be enabled or disabled as well

tmp zflxpjcth5_screenshot

Let me know if something need to be changed

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -97,7 +97,7 @@ if [ -n "$tag_filter" ]; then
       TAGS=$(git tag --list "$tag_filter" --sort=creatordate --contains $ref)
       get_commit $TAGS
     elif [ -n "$ref" ]; then
-      TAGS=$(git tag --list "$tag_filter" --list --sort=creatordate | lines_including_and_after $ref)
+      TAGS=$(git tag --list "$tag_filter" --sort=creatordate | lines_including_and_after $ref)
       get_commit $TAGS
     else
       TAG=$(git tag --list "$tag_filter" --sort=creatordate | tail -1)
--- a/assets/check
+++ b/assets/check
@@ -34,7 +34,7 @@ configure_git_global "${git_config_payload}"
 destination=$TMPDIR/git-resource-repo-cache
 
 tagflag=""
-if [ -d $tag_filter ]; then
+if [ -n "$tag_filter" ]; then
   tagflag="--tags"
 fi
 
--- a/assets/check
+++ b/assets/check
@@ -85,23 +85,23 @@ lines_including_and_after() {
 }
 
 get_commit(){
-  for TAG in $*; do
-    COMMIT=$(git rev-list -n 1 $TAG)
-    echo "{ \"ref\": \"$TAG\", \"commit\": \"$COMMIT\" }"
+  for tag in $*; do
+    commit=$(git rev-list -n 1 $tag)
+    jq -n '{ref: $tag, commit: $commit}' --arg tag $tag --arg commit $commit
   done
 }
 
 if [ -n "$tag_filter" ]; then
   {
     if [ -n "$ref" ] && [ -n "$branch" ]; then
-      TAGS=$(git tag --list "$tag_filter" --sort=creatordate --contains $ref)
-      get_commit $TAGS
+      tags=$(git tag --list "$tag_filter" --sort=creatordate --contains $ref)
+      get_commit $tags
     elif [ -n "$ref" ]; then
-      TAGS=$(git tag --list "$tag_filter" --sort=creatordate | lines_including_and_after $ref)
-      get_commit $TAGS
+      tags=$(git tag --list "$tag_filter" --sort=creatordate | lines_including_and_after $ref)
+      get_commit $tags
     else
-      TAG=$(git tag --list "$tag_filter" --sort=creatordate | tail -1)
-      get_commit $TAG
+      tag=$(git tag --list "$tag_filter" --sort=creatordate | tail -1)
+      get_commit $tag
     fi
   } | jq -s "map(.)" >&3
 else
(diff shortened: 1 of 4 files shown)


vito:
This seems to overlap with https://github.com/concourse/git-resource/pull/225 which was submitted later but seems a bit more thorough. Do you mind if we just focus our efforts on that PR?

Thanks for submitting though!

talset:
@vito thx for your answer. Well as you want, we can fix the --tags part on #225 (I going to comment few question/changes) and them "rebase" this PR regarding "Adding commit id with ref in order to be able to override tag

goddenrich:
Nice I like this too! was thinking it would be great to see the commit as well as the tag.

talset:
@vito 

> This seems to overlap with #225 which was submitted later but seems a bit more thorough. Do you mind if we just focus our efforts on that PR?

After verification it would have been easier to continue on this one. The only 2 things missing are the unit test and the line git with sed. But it is you call ;)

My suggestion have been added on #225

vito:
@talset Thanks for your help on the other PR! Now that I look at it these two have different end goals so that would explain the difference in testing. Didn't mean to imply theirs was particularly better or anything. :slightly_smiling_face: But at least we can get the other one in first to reduce merge conflicts.

talset:
Thanks @vito, I rebased my commit and keep the only missing part to add commit and ref when we are using tags.

Let me know if something need to be changed.

assets/check
```diff
@@ -34,7 +34,7 @@ configure_git_global "${git_config_payload}"
 destination=$TMPDIR/git-resource-repo-cache
 
 tagflag=""
-if [ -d $tag_filter ]; then
```
vito: Bleh, that looks broken, thanks for noticing. I'm gonna fix that on `master` just to unbreak things independent of this PR. I wonder why the tests didn't catch this?

talset: well this condition is only used to add a new `--tags flag` on git (the magic to detect tags on different branch). But the test are only done on one branch (branch-a) so it have no impact on the result. I not check how it could be done but the idea would be

`make a commit + tag on branch-b, clone --single-branch branch-a and look at tags on branch-b`

goddenrich: My bad! I'll look into writing some tests to cover this

talset: Im not sure how you will able to cover it with local branch creation because --tags is used to get the remote branch from a clone --single-branch. But localy you will already have all the branchs.

Well hope you will find a tricks :)

assets/check
```diff
@@ -84,16 +84,26 @@ lines_including_and_after() {
   sed -ne "/$escaped_string/,$ p"
 }
 
+get_commit(){
+  for TAG in $*; do
+    COMMIT=$(git rev-list -n 1 $TAG)
+    echo "{ \"ref\": \"$TAG\", \"commit\": \"$COMMIT\" }"
```
vito: I would suggest using `jq` to build the JSON using `--arg`, just in case someone has funny tag names:

```sh
jq -n '{ref: $tag, commit: $commit}' --arg tag $TAG --arg commit $COMMIT
```

Minor nit: we've been using lowercase for variable names since it's a bit harder to confuse them with global env vars.

talset: Thanks @vito
I updated the PR regarding using `jq` and lowercase variable

assets/in
```diff
@@ -69,11 +69,11 @@ if test "$depth" -gt 0 2> /dev/null; then
 fi
 
 tagflag=""
-if [ -d $tag_filter ]; then
+if [ -n "$tag_filter" ]; then
   tagflag="--tags"
 fi
 
-git clone $depthflag $uri $branchflag $destination $tagflag
+git clone --single-branch $depthflag $uri $branchflag $destination $tagflag
```
vito: note to future self: this just undoes part of https://github.com/concourse/git-resource/commit/a62caf6dfdc8bd5f5bd579d85b48817cd980145e since it wasn't necessary (comment here: https://github.com/concourse/git-resource/pull/225#discussion_r234826904)

assets/check
prein: Thanks @vito and @talset

michaelklishin:
Any chance for a new release that would include this?

talset:
> Any chance for a new release that would include this?

I don't know when it will be realesed on the official docker hub but If you need it temporarly you can use mine `cycloid/concourse-git-resource:latest`

vito:
@michaelklishin @talset Just published 1.1, sorry for the delay! It's available on Docker Hub as `concourse/git-resource:1.1` (and latest/1/1.1.0).

lancempoe:
Will this fix the original problem, https://github.com/concourse/git-resource/issues/74, that all tags should trigger, and ignore branches completely?

talset:
This PR fix only the ability to replace an existing tag but the previous PR with 

```
if [ -n "$tag_filter" ]; then
  tagflag="--tags"
fi
```

do the job of ignore branches  for tag_filter

PR #226, opened 2018-11-21
Fix submodule deepening

norbertbuchmueller:
The fix for deepening submodules in #212 introduced two bugs:

* When `params.depth` was not set (or was set to 0) it failed to update (check out) the submodules, see https://github.com/concourse/git-resource/pull/212#issuecomment-440473385
* The code that restored the submodule update method was broken, see https://github.com/concourse/git-resource/pull/212#issuecomment-440501169

This PR fixes those (plus an unrelated ancient bug in a test) and improves the tests of checking out submodules.

### DIFF
--- a/assets/deepen_shallow_clone_until_ref_is_found
+++ b/assets/deepen_shallow_clone_until_ref_is_found
@@ -20,31 +20,30 @@ readonly ref="$1"
 # In either case we try to deepen the shallow clone until we find $ref, reach
 # the max depth of the repo, or give up after a given depth and resort to deep
 # clone.
-if [ "$concourse_git_resource__depth" -gt 0 ]; then
-  depth="$concourse_git_resource__depth"
-  git_dir="$(git rev-parse --git-dir)"
-  readonly git_dir
 
-  while ! git checkout -q "$ref" &>/dev/null; do
-    # once the depth of a shallow clone reaches the max depth of the origin
-    # repo, Git silenty turns it into a deep clone
-    if [ ! -e "$git_dir"/shallow ]; then
-      echo "Reached max depth of the origin repo while deepening the shallow clone, it's a deep clone now"
-      break
-    fi
+depth="$concourse_git_resource__depth"
+git_dir="$(git rev-parse --git-dir)"
+readonly git_dir
 
-    echo "Could not find ref ${ref} in a shallow clone of depth ${depth}"
+while ! git checkout -q "$ref" &>/dev/null; do
+  # once the depth of a shallow clone reaches the max depth of the origin
+  # repo, Git silenty turns it into a deep clone
+  if [ ! -e "$git_dir"/shallow ]; then
+    echo "Reached max depth of the origin repo while deepening the shallow clone, it's a deep clone now"
+    break
+  fi
 
-    (( depth *= 2 ))
+  echo "Could not find ref ${ref} in a shallow clone of depth ${depth}"
 
-    if [ "$depth" -gt "$max_depth" ]; then
-      echo "Reached depth threshold ${max_depth}, falling back to deep clone..."
-      git fetch --unshallow origin
+  (( depth *= 2 ))
 
-      break
-    fi
+  if [ "$depth" -gt "$max_depth" ]; then
+    echo "Reached depth threshold ${max_depth}, falling back to deep clone..."
+    git fetch --unshallow origin
 
-    echo "Deepening the shallow clone to depth ${depth}..."
-    git fetch --depth "$depth" origin
-  done
-fi
+    break
+  fi
+
+  echo "Deepening the shallow clone to depth ${depth}..."
+  git fetch --depth "$depth" origin
+done
--- a/assets/deepen_shallow_clone_until_ref_is_found_then_check_out
+++ b/assets/deepen_shallow_clone_until_ref_is_found_then_check_out
@@ -5,9 +5,8 @@ set -e
 
 readonly max_depth=128
 
-readonly ref="$1"
-
-# the concourse_git_resource__depth variable is exported by the 'in' script
+declare depth="$1"
+readonly ref="$2"
 
 # A shallow clone may not contain the Git commit $ref:
 # 1. The depth of the shallow clone is measured backwards from the latest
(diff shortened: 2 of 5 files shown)


assets/deepen_shallow_clone_until_ref_is_found
```diff
@@ -20,31 +20,30 @@ readonly ref="$1"
...
+  fi
+
+  echo "Could not find ref ${ref} in a shallow clone of depth ${depth}"
+
+  (( depth *= 2 ))
+
+  if [ "$depth" -gt "$max_depth" ]; then
+    echo "Reached depth threshold ${max_depth}, falling back to deep clone..."
+    git fetch --unshallow origin
+
```
ndmckinley: You're missing `git checkout -q $ref` here - you'll unshallow the repo, but not check out the right commit, if you're in a submodule.

assets/deepen_shallow_clone_until_ref_is_found
```diff
@@ -20,31 +20,30 @@ readonly ref="$1"
...
+depth="$concourse_git_resource__depth"
+git_dir="$(git rev-parse --git-dir)"
+readonly git_dir
+
+while ! git checkout -q "$ref" &>/dev/null; do
+  # once the depth of a shallow clone reaches the max depth of the origin
+  # repo, Git silenty turns it into a deep clone
+  if [ ! -e "$git_dir"/shallow ]; then
+    echo "Reached max depth of the origin repo while deepening the shallow clone, it's a deep clone now"
+    break
```
ndmckinley: This is an error state - you weren't able to check out the ref, but there's no further you can search.  You should crash.

Note: this will crash correctly for the main checkout (because there's an extra `git checkout` in the main file, which, if it does not succeed, will crash because of `set -e`), but it will fail for submodules because the submodules do not have that extra `git checkout`.

assets/in
```diff
@@ -73,8 +73,10 @@ cd $destination
...
 git fetch origin refs/notes/*:refs/notes/*
 
-concourse_git_resource__depth="$depth" \
-  "$bin_dir"/deepen_shallow_clone_until_ref_is_found "$ref"
+if [ "$depth" -gt 0 ]; then
+  concourse_git_resource__depth="$depth" \
+    "$bin_dir"/deepen_shallow_clone_until_ref_is_found "$ref"
+fi
 
 git checkout -q $ref
```
ndmckinley: This belongs in an `else` for the above `if` - it's not an error, exactly, but it's superfluous.

assets/in
```diff
@@ -134,26 +136,28 @@ if [ "$submodules" != "none" ]; then
...
-    if [ "$update_conf_was_set" != 0 ]; then
-      git config "submodule.${submodule}.update" "$update_conf"
-    else
-      git config --unset "submodule.${submodule}.update"
-    fi
-  done
+  if [ "$depth" -gt 0 ]; then
+    for submodule in $submodules; do
+      # remember submodule update config
+      update_conf_was_set=0
```
ndmckinley: Since `""` is an invalid value for `.update`, you might just use `[ -z $update_conf ]` below when deciding whether to set it, instead of this.

vito:
@norbertbuchmueller Thanks for following up on this! Do you have a chance to review the above feedback?

assets/deepen_shallow_clone_until_ref_is_found
norbertbuchmueller: Indeed. Fixed in 82ef5a393a1b1e5a00f5413175f10350800491e8 and ad553ecb6aa971840fb6be0b805a50616d2b2480.

assets/deepen_shallow_clone_until_ref_is_found
norbertbuchmueller: Added a test in e31d5cea9237e7d833dc4d18709df822814e97ad and ad553ecb6aa971840fb6be0b805a50616d2b2480 fixes it.

assets/in
norbertbuchmueller: Fixed in ad553ecb6aa971840fb6be0b805a50616d2b2480.

assets/in
norbertbuchmueller: Yes, you're right that that would work and it would be a bit simpler. Actually i considered that approach when i wrote the code originally, and thought that later Git may allow `""` as a valid value for `.update` and then that would break the code. So i thought it's better this way.

norbertbuchmueller:
@ndmckinley Thank you very much for the review! If you could have another look to verify my fixes that would be really great.

@vito Sorry about the long delay.

assets/deepen_shallow_clone_until_ref_is_found_then_check_out
```diff
@@ -0,0 +1,50 @@
...
+#    $ref that are not affecting the filtered paths.
+#
+# In either case we try to deepen the shallow clone until we find $ref, reach
+# the max depth of the repo, or give up after a given depth and resort to deep
+# clone.
+
+git_dir="$(git rev-parse --git-dir)"
+readonly git_dir
+
+while ! git checkout -q "$ref" &>/dev/null; do
```
ndmckinley: If it were me, I would change this `checkout` to be one of the different commands for checking if a commit exists.  Now that the checkout happens at the end, you don't need it here, and it's mildly clearer for an if test not to have side effects.

I prefer `git cat-file -e "$ref"` because it's branch independent (and -e means "exit 0 if valid, exit 1 if not"), but you can also use `git rev-list "HEAD..$ref"` or I've actually seen people advocate for `git merge-base "$ref" "$ref"`, which I find kind of goofy.

assets/in
ndmckinley: If you ask me I think that's extremely unlikely - I doubt they'll add a recognition of the difference between "set to `""`" and "unset".  But I see you're going for future-proofing ... I have no opinion, then.  :)

vito:
@ndmckinley Thanks for reviewing! I think this is good to go too, so I'll merge this in to fix those bugs and ship a new version so y'all can test it out. Happy to review any further PRs!

@norbertbuchmueller thanks again!

norbertbuchmueller:
@ndmckinley Thanks for the review!

> I'll merge this in to fix those bugs and ship a new version so y'all can test it out.

Thanks @vito! Looking forward to try the next release.

PR #234, opened 2019-01-04
fix #233 error would clobber existing tag

### LINKED ISSUE #233: [replace tags] error would clobber existing tag
By using tag and replacing one with `git tag -f ` raise the following git error

```
From github.com:talset/crappytest
 ! [rejected]        DEV        -> DEV  (would clobber existing tag)
```

How to reproduce:
```
git checkout bla
git tag -f DEV
git push origin DEV --force

git checkout foo
git tag -f DEV
git push origin DEV --force
```

What is done by the resource ?

```
git fetch --tags
From github.com:talset/crappytest
 ! [rejected]        DEV        -> DEV  (would clobber existing tag)
```

How to solve it ?

Add a `-f` flag
```
git fetch --tags -f
From github.com:talset/crappytest
 t [tag update]      DEV        -> DEV
```

I will make a PR for this

talset:
issue: #233

Update git fetch with `-f`

```
  -f, --force           force overwrite of local reference
```

ensure local refs are always the same as the upstream

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -518,6 +518,24 @@ it_can_check_with_tag_filter_with_bogus_ref() {
   "
 }
 
+it_can_check_with_tag_filter_with_replaced_tags() {
+
+  local repo=$(init_repo)
+  local ref1=$(make_commit_to_branch $repo branch-a)
+  local ref2=$(make_annotated_tag $repo "staging" "tag branch-a")
+  # see that the tag is initially ref1
+  check_uri_with_tag_filter $repo "staging" | jq -e "
+    . == [{ref: \"staging\", commit: \"$ref1\"}]
+  "
+
+  local ref3=$(make_commit_to_branch $repo branch-a)
+  local ref4=$(make_annotated_tag $repo "staging" "tag branch-a")
+
+  check_uri_with_tag_filter $repo "staging" | jq -e "
+    . == [{ref: \"staging\", commit: \"$ref3\"}]
+  "
+}
+
 it_can_check_and_set_git_config() {
   local repo=$(init_repo)
   local ref=$(make_commit $repo)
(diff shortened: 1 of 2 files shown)


vito:
Thanks! Could a test be added for this? i.e. use local repo with tags as `uri:`, run check, update tag in repo, see that tag is updated during fetch?

talset:
> Thanks! Could a test be added for this? i.e. use local repo with tags as `uri:`, run check, update tag in repo, see that tag is updated during fetch?

I added a new test function `it_can_check_with_tag_filter_with_replaced_tags_from_cache` who create a copy of the repo in order to use it as `git-resource-repo-cache` with an outdated version. It allow us to do a `git fetch --tags` from a remote origin.

Let me know if something need to be improved

test/check.sh
```diff
@@ -518,6 +518,20 @@ it_can_check_with_tag_filter_with_bogus_ref() {
...
+it_can_check_with_tag_filter_with_replaced_tags_from_cache() {
+
+  local repo=$(init_repo)
+  local ref1=$(make_commit_to_branch $repo branch-a)
+  local ref2=$(make_annotated_tag $repo "staging" "tag branch-a")
+  local cache_repo=$(make_cache_repo $repo)
+  local ref3=$(make_commit_to_branch $repo branch-a)
+  local ref4=$(make_annotated_tag $repo "staging" "tag branch-a")
+
+  check_uri_with_tag_filter $cache_repo "staging" | jq -e "
```
vito: I wonder if you could avoid the `make_cache_repo` by just running the check twice?

Something like:

```sh
  local repo=$(init_repo)
  local ref1=$(make_commit_to_branch $repo branch-a)
  local ref2=$(make_annotated_tag $repo "staging" "tag branch-a")

  # see that the tag is initially ref2
  check_uri_with_tag_filter $cache_repo "staging" | jq -e "
    . == [{ref: \"staging\", commit: \"$ref2\"}]
  "

  local ref3=$(make_commit_to_branch $repo branch-a)
  local ref4=$(make_annotated_tag $repo "staging" "tag branch-a")

  # see that it is now ref3
  check_uri_with_tag_filter $cache_repo "staging" | jq -e "
    . == [{ref: \"staging\", commit: \"$ref3\"}]
  "
```

IMO this is a bit cleaner and more black-box style. wdyt?

talset: I didn't think to run it twice. Agree it should be cleaner. Im going to do it

talset: fixed

vito:
awesome, thanks a bunch!

PR #237, opened 2019-02-02
Allow specifying submodule_recursive in source config (take 2).

alext:
This is a reworking of #210 to fix issues raised with it. Apologies for the delay in responding to comments (I realised that I've now force-pushed to the branch, so I don't think the original PR can be reopened - hence opening a new one). I realise that there's an upcoming reworking of resources in general, but this would still be very useful or us in the meantime.

## Original description

This adds the option for specifying submodule_recursive in the source
definition. This avoids the need to specify this on every get for a
given git resource in the pipeline. If this is specified in both places, the
get value will take precedence.

While testing this behaviour, I've also noticed that it's necessary to specify
this param as a string due to the way that the default value is implemented
using jq. I've added a note to the docs to clarify this.


vito:
@alext Hey, sorry for the delay! Would you mind rebasing these commits with `-s` to sign off on the DCO? Instructions are here: https://github.com/concourse/git-resource/pull/237/checks?check_run_id=58073512

More information here if you're curious: https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#signing-your-work

alext:
@vito Hi, I've rebased as requested with the sign-off comments.

test/get.sh
```diff
@@ -410,18 +410,18 @@ it_honors_the_parameter_flags_for_submodules() {
   "file://"$project_folder 1 "all" true $dest_recursive_true |  jq -e "
     .version == {ref: $(echo $project_last_commit_id | jq -R .)}
   "
-  test "$(git -C $project_folder rev-parse HEAD)" = $project_last_commit_id
-  test "$(git -C $dest_recursive_true/$submodule_name rev-list --all --count)" = 1
-  test "$(git -C $dest_recursive_true/$submodule_name/$subsubmodule_name rev-list --all --count)" = 1
+  test "$(git -C $dest_recursive_true rev-parse HEAD)" = $project_last_commit_id
```
vito: Was this assertion broken before? Looks like it was testing against the source repo rather than the fetched repo?

test/get.sh
```diff
@@ -410,18 +410,18 @@ it_honors_the_parameter_flags_for_submodules() {
...
+  test "$(ls $dest_recursive_true/$submodule_name | wc -l)" = 2
+  test "$(ls $dest_recursive_true/$submodule_name/$subsubmodule_name | wc -l)" = 1
```
vito: What're these changes for? Just curious.

test/get.sh
alext: Yep, this looked to be testing the wrong repo before.

test/get.sh
alext: Hi, there are some details in the commit message for a95a98a, but essentially, this updates these to test the behaviour that differs around the submodule_recursive flag, which is the number of files present as opposed to the commit depth. It looks as though the original assertions may have been copied from the tests for the depth flag.

vito: Ah, thanks for the descriptive commits!

vito:
thanks!

vito:
Hey @alext, totally my bad but the tests are actually failing. I saw green status checks and hadn't noticed it was actually because the tests hadn't even run yet, and it was only the DCO/WIP checks that had passed. I'm going to revert this since `master` now has failing tests. If you want to recover the changes you can just revert my revert and work from there.

Sorry for the mess - happy to review again if you want to PR it back!

PR #121, opened 2017-05-31
Allow the use of ProxyCommand

robbruce:
Allow the use of ProxyCommand which covers more usecases beyond corkscrew, using nc for example. Fixes concourse/git-resource#84

PR #238, opened 2019-02-04
Use OpenBSD netcat instead of busybox netcat

drennalls:
I have a setup where a particular git resource can only be accessed via a SOCKS proxy (created using `ssh -D <port>`). Git can be configured to go through a SOCKS proxy via netcat/nc by setting the [`core.sshCommand` option](https://git-scm.com/docs/git-config#git-config-coresshCommand) to something like `ssh -o ProxyCommand="nc -X 5 -x <proxy host/ip>:<proxy port> %h %p"`. Configuring the git resource to achieve the same thing can be done using the resource's `git_config` option like this:
```
  git_config:
  - name: core.sshCommand                                                                                                                                                           
    value: ssh -o ProxyCommand="nc -X 5 -x <proxy host/ip>:<proxy port> %h %p" 
```
However because the netcat/`nc` binary that's currently bundled (from Busybox) does not include the proxy support options doing this actually fails. By simply pulling in the OpenBSD version of netcat I was able to have things work as expected.

This would address https://github.com/concourse/git-resource/pull/121 as well

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN apk --no-cache add \
   libressl-dev \
   make \
   g++ \
+  netcat-openbsd \
   openssh \
   perl \
   tar \


drennalls:
FWIW in terms of testing, I've built a Docker image with this change, pushed it my dockerhub account and am using it locally without issues
..pipeline resource definition...
```
 resource_types:                                                                                                                                                                                                                                                                 
- name: git-with-netcat                                                                                                                                                                                                                                                         
  type: docker-image                                                                                                                                                                                                                                                            
  source:                                                                                                                                                                                                                                                                       
    repository: drennalls/git-resource                                                                                                                                                                                                                                          
    tag: latest                    
```
..usage..
```
- name: my-git-repo                                                                                                                                                                                                                                                   
  type: git-with-netcat                                                                                                                                                                                                                                                         
  source:                                                                                                                                                                                                                                                                       
    uri: ((config_repo_source_uri))                                                                                                                                                                                                                                             
    branch: ((config_repo_input_branch_name))                                                                                                                                                                                                                                   
    private_key: ((concourse_ci_private_key))
  git_config:
  -  name: core.sshCommand                                                                                                                                                                                                                                                       
     value: ssh -o ProxyCommand="nc -X 5 -x 192.168.0.107:9000 %h %p"

vito:
Gonna close this since it's gotten a bit stale, but happy to keep discussing in the future.

PR #239, opened 2019-02-05
in: Always fetch tags

giner:
*Why*
- When pipeline is triggered on a particular branch it is useful to have
all tags being accessible including those which don't belong to the
cloned branch. For instance this allows to detect if the branch is
rebased on the latest release.

*Changes*
- remove optional parameter `--tags` from git `git clone` as there is no such a flag
- add non-optional `--tags` to `fetch origin ...`

Signed-off-by: Stanislav German-Evtushenko <s.germanevtushenko@rakuten.com>

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -77,7 +77,7 @@ git clone --single-branch $depthflag $uri $branchflag $destination
 
 cd $destination
 
-git fetch origin refs/notes/*:refs/notes/* --tags
+git fetch origin refs/notes/*:refs/notes/* $( [ "$depth" -gt 0 ] && echo "$tagflag" || echo "--tags" )
 
 if [ "$depth" -gt 0 ]; then
   "$bin_dir"/deepen_shallow_clone_until_ref_is_found_then_check_out "$depth" "$ref" "$tagflag"
--- a/assets/in
+++ b/assets/in
@@ -73,11 +73,11 @@ if [ -n "$tag_filter" ]; then
   tagflag="--tags"
 fi
 
-git clone --single-branch $depthflag $uri $branchflag $destination $tagflag
+git clone --single-branch $depthflag $uri $branchflag $destination
 
 cd $destination
 
-git fetch origin refs/notes/*:refs/notes/* $tagflag
+git fetch origin refs/notes/*:refs/notes/* --tags
 
 if [ "$depth" -gt 0 ]; then
   "$bin_dir"/deepen_shallow_clone_until_ref_is_found_then_check_out "$depth" "$ref" "$tagflag"


vito:
@giner One problem is the state of tags is going to be cached. That is, if your pipeline ever fetches commit `abcdef`, all later runs will only see the tags present at the time it was initially cloned. Tags created later won't be present for later jobs/builds that run on the same worker.

This change might be fine for some use cases, i.e. if you know by the time you fetch a commit you'll have all the tags you're checking for, but I'm hesitant to let people rely on this as it could lead to gotchas later on 
in their pipeline.

For this reason I usually recommend people use a separate resource that's explicitly keeping track of tags in the repo. You could use another `git` resource with `tag_filter` for example. Would this work for you?

giner:
I'm not sure I'm getting the point. `git clone` fetches tags anyway unless `--no-tags` is specified (see below) which is not the case. The problem is that with `--single-branch` it fetches only tags which point to commits on the specified branch but I need all of them to be available.
```
$ TMPDIR=$(mktemp -d)
$ git clone --single-branch https://github.com/concourse/git-resource --branch=xenial-base-image $TMPDIR

Cloning into '/var/folders/cc/7_t7fcj56pz3zb0m6mj93911jhxb8x/T/tmp.pEcWlY9i'...
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 1453 (delta 0), reused 1 (delta 0), pack-reused 1448
Receiving objects: 100% (1453/1453), 318.62 KiB | 0 bytes/s, done.
Resolving deltas: 100% (849/849), done.
Checking connectivity... done.

$ git -C $TMPDIR tag

v1.0.0
v1.0.1
v1.1.0
v1.1.1
```

> For this reason I usually recommend people use a separate resource that's explicitly keeping track of tags in the repo

What I need is to be able to detect the latest release (by tags) and compare HEAD of the branch which triggered pipeline to the latest commit, e.g. `git log --oneline --left-right "v1.1.1"...`.

vito:
@giner The resource only fetches tags right now because it was too late to remove that functionality. Someone actually PR'd a change to prune all tags but we had to roll it back as too many people were still depending on it.

My point is only that tag state is prone to becoming out of date as the resource caches will only have whichever tags were present at the time. If new tags are created later (i.e. a new release), any builds that run after that and use the cache won't see the new tag. So relying on tags that are implicitly fetched and cached by the resource is risky.

> What I need is to be able to detect the latest release (by tags) and compare HEAD of the branch which triggered pipeline to the latest commit, e.g. git log --oneline --left-right "v1.1.1"....

Makes sense, just be aware that if that build ever runs again with the same commit, and a new version has shipped since then (say `v1.1.2`), it won't have that tag. Using a separate resource to explicitly track the tagged versions would fix this:

```yaml
resources:
- name: my-repo-dev
  type: git
  source:
    uri: https://github.com/concourse/concourse
    branch: master

- name: my-repo-released
  type: git
  source:
    uri: https://github.com/concourse/concourse
    tag_filter: 'v*'

jobs:
- name: compare
  plan:
  - get: my-repo-dev
  - get: my-repo-released
  - task: compare
    # has both as inputs
```

This works because `tag_filter:` looks across all branches and is versioned based on the tags themselves. So it won't be out of date and all caching will work fine.

giner:
I think there is some confusing here. I don't need these tags to be in versions, pipeline is triggered on a branch anyway. What I need is just the repo to be cloned with tags so that I could run "git tags" later and get the latest one.

Btw, I have just realized that with `depth` specified the behavior changes if the PR is merged. Though it's easy to fix.

vito:
I get what you're trying to do, I'm saying that by by doing it this way you're setting yourself up for a gotcha due to Concourse's resource caching semantics, because the tags aren't technically part of the input (only the commit is).

This is the scenario:

1. a build runs with commit `abcdef` - latest tag is currently `v1.1.1`
    * runs `git log --oneline --left-right "v1.1.1"...`
1. a new version is shipped and tagged as `v2.0.0` (or whatever)
1. a build runs again with commit `abcdef`, but this has already been fetched and cached, and the cache only has `v1.1.1`
    * so it still runs `git log --oneline --left-right "v1.1.1"...`

This might be fine for your use case, but I would still recommend the other method as it avoids pointing a loaded gun at your foot. I don't see how it makes it that much more complicated, and if anything it feels clearer to me - you're comparing to the latest shipped version, so shouldn't that be an input to the build? I don't know how many jobs you have that do this, so maybe it is, but that seems like a small price to pay for correctness. :man_shrugging:

giner:
Now I get it. I was thinking about versions and didn't realize that resources are cached 🤦‍♂️ Shooting in one's own foot appeared to be easier than I could think.

Issue #240, opened 2019-02-06
switch to using ubuntu for base image


chenbh:
## Why
tldr: Patching ubuntu images is faster and more reliable than alpine

## Impacts
* The resource image will be larger (272mb -> 383mb)

larssb:
Hi @chenbh,

Out of curiosity I would like to know/read info on `alpine` being less reliable in regards to patching and day-to-day operation. Do you have some URI's or other types of documentation on this?

I would highly appreciate it.

Thank you and have a good 1.

chenbh:
Hey @larssb there's more context in the parent issue https://github.com/concourse/concourse/issues/3266#issuecomment-463227274

One more thing to note is that we are currently using `alpine:edge` which can be a bit unstable due to its nature. 

If we are to switch to Alpine 3.9 (latest stable currently) it still only has a support window of 2 years till 2021 (https://wiki.alpinelinux.org/wiki/Alpine_Linux:Releases).

On the other hand Ubuntu 18.04 has a LTS till 2023 (https://wiki.ubuntu.com/Releases)

PR #241, opened 2019-02-06
use ubuntu for base image

chenbh:
concourse/git-resource#240
part of concourse/concourse#3266

* remove `proxytunnel` intermediate layer in favor of building it in the final layer directly
* various `alpine` to `ubuntu` dependency compatibility changes
* switch to cloning `git-crypt` repo and using tags instead of downloading and untar-ing it
* `git-lfs` does not have an ubuntu distro, we are currently running some arbitrary script off the site to install it which is pretty sketchy


chenbh:
Still working on dealing with `git-lfs` so this is not completely ready to merge yet

dockerfiles/ubuntu/Dockerfile
```diff
@@ -0,0 +1,195 @@
...
+  libstdc++6 \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
+RUN apt-get install -y git-lfs
+
+WORKDIR /root
+RUN git clone https://github.com/proxytunnel/proxytunnel.git; \
+    cd proxytunnel; \
+    make; \
```
cirocosta: A little optimization (assuming most of the times the build will be built in a machine that has multiple cores available), as their build makes use of a list of dependencies https://github.com/proxytunnel/proxytunnel/blob/f7831a292f7182dc20e5dbd11a90c18474b6077b/Makefile#L82-L84, we can specify a concurrency factor to `make` to speed the build up with - `make -j4` would do it

dockerfiles/ubuntu/Dockerfile
```diff
@@ -0,0 +1,195 @@
...
+  openssl \
+  libssl-dev \
+  make \
+  g++ \
+  openssh-client \
+  libstdc++6 \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash
+RUN apt-get install -y git-lfs
```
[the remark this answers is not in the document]
YoussB: yes, since we are using a shell script to get the package repo of `git-lfs` as it is not included in the official ubuntu repo.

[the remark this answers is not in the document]
chenbh: `git-lfs` is hosted on it's own package repo that the line 18 sets up (yeah it's kinda iffy we're running some arbitrary bash script from the internet but this is the "official" way).

It's separated because we need to install curl to download the setup script

README.md
```diff
@@ -252,7 +252,7 @@ will stop the build.
 Run the tests with the following command:
 
 ```sh
-docker build -t git-resource .
+docker build -t git-resource -f ./dockerfiles/ubuntu/Dockerfile .
```
cirocosta: Hey, as we're not dropping the support for `alpine` in this PR, I'd say we should advertise the fact that it's possible to build and test both.

Wdyt?

thanks!

chenbh: I would assume that since the path is provided as `dockerfiles/.../Dockerfile` it would be intuitive enough to check that directory and see that there are multiple options.

Although now that I think about it, the README should use `alpine` as that's still the default option for now

cirocosta: Sounds good 👍

chenbh:
@cirocosta we also applied the optimization changes to the `alpine` dockerfile as well

PR #242, opened 2019-02-08
Create a git tag-like ref

SyrianSpock:
This PR uses `git describe --tags --always` to generate a tag-like ref.
It allows us to get a unique version describing the current HEAD of the git resource. This has a nice
semver format that can be used in a pipeline for versioning artifacts and releases.
Note that `--always` makes the command fail-proof as it falls back to a short commit SHA-1 ref if the repo was never tagged.

### DIFF
--- a/test/get.sh
+++ b/test/get.sh
@@ -646,6 +646,10 @@ it_can_get_returned_ref() {
   test "$(cat $dest/.git/ref)" = "${ref3}" || \
     ( echo ".git/ref does not match. Expected '${ref3}', got '$(cat $dest/.git/ref)'"; return 1 )
 
+  test -e $dest/.git/tag_ref || ( echo ".git/tag_ref does not exist."; return 1 )
+  test "$(cat $dest/.git/tag_ref)" = "${ref3}" || \
+    ( echo ".git/tag_ref does not match. Expected '${ref3}', got '$(cat $dest/.git/tag_ref)'"; return 1 )
+
   test -e $dest/.git/short_ref || ( echo ".git/short_ref does not exist."; return 1 )
   local expected_short_ref="test-$(echo ${ref3} | cut -c1-7)"
   test "$(cat $dest/.git/short_ref)" = $expected_short_ref || \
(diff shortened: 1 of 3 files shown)


SyrianSpock:
@vito any feedback on this PR would be greatly appreciated, thanks.

vito:
@SyrianSpock Sorry, I held off on PRs for a bit to focus on getting 5.0 out. I'll be able to review this soon.

danilopopeye:
@SyrianSpock the `--tags` matches non annotated tags also. [1]

Maybe we could have a option to only return annotated tags ?

> [1] [--tags](https://git-scm.com/docs/git-describe#Documentation/git-describe.txt---tags)
> Instead of using only the annotated tags, use any tag found in refs/tags namespace. This option enables matching a lightweight (non-annotated) tag.

SyrianSpock:
ping @vito

test/get.sh
```diff
@@ -646,6 +646,10 @@ it_can_get_returned_ref() {
   test "$(cat $dest/.git/ref)" = "${ref3}" || \
     ( echo ".git/ref does not match. Expected '${ref3}', got '$(cat $dest/.git/ref)'"; return 1 )
 
+  test -e $dest/.git/tag_ref || ( echo ".git/tag_ref does not exist."; return 1 )
+  test "$(cat $dest/.git/tag_ref)" = "${ref3}" || \
```
vito: Unless I'm missing something, this doesn't seem to test that we actually output `git describe` as it ends up just being a regular ol' commit. Could we add a test that sets up tags and such and asserts against the specific tag name?

deadNightTiger:
@SyrianSpock Are you planning to finish this PR? I'm really interested in this and could take it over in case you are no longer interested.

spikewang:
This will be really useful. I have to implement a resource type to achieve this... It will be great if this could get in for the next release...

wagdav:
Hello all,
I'm a co-worker of @SyrianSpock. @SyrianSpock took some time off from coding so work won't continue on this PR in the near future.

@vito  Would you please close this PR and take a look at #281 ? Thanks!

vito:
@wagdav :turtle: closing!

PR #245, opened 2019-03-03
fix `test/get.sh tests

danilopopeye:
I've cloned this repo to work in some feature and noted some tests were failing.

I think I fixed them 😄


vito:
Woops - I've reverted the change that broke master. I'm not sure if the fixes proposed here are correct; seems safer to just revert for now.

More context: https://github.com/concourse/git-resource/pull/237#issuecomment-469911141

Thanks though!

danilopopeye:
no problem @vito, the main point was to get things working 😉

PR #247, opened 2019-03-13
fix discrepancy between name/path with submodules

### LINKED ISSUE #246: Submodule update uses name instead of path for submodules: all
Hi!

I'm using Concourse to build/deploy my applications. I upgraded my Concourse BOSH deployment from v4.2.3 to v5.0.0 and found an issue when my pipeline pulls my project repo.

### Issue
Repositories containing submodules with names that are not the same as their path cannot be retrieved through this resource. The submodule update fails because the name is used rather than the path.

Here's a quick way to reproduce the issue:

Create a new git repository and add a submodule with the following command:
`git submodule add --name "testmodule" <submodule_repository_url> "path/to/submodule"`

This should create the following .gitmodules:
```
[submodule "testmodule"]
	path = path/to/submodule
	url = <submodule_repository_url>
```
Commit & push these changes to a remote repo. Create a pipeline that uses this resource to pull from the repo with default `get` params (or `submodules: all`).

The `get` step fails in Concourse v5.0.0 with output:
```
Submodule 'testmodule' (<submodule_repository_url>) registered for path 'path/to/submodule'
error: pathspec 'testmodule' did not match any file(s) known to git
```

It looks to me that the issue comes from [this](https://github.com/concourse/git-resource/blob/62f1204144b2e7c37a35bae1b322bf82491b8745/assets/in#L137) method of determining the submodules, which occurs if `submodules: all` param is used for the `get` step. Using `git config` lists the submodule names, where the `git submodule update` [further down](https://github.com/concourse/git-resource/blob/62f1204144b2e7c37a35bae1b322bf82491b8745/assets/in#L156) expects paths to the submodules.

I think using `git submodule` and extracting the path from the output should work.

A quick fix for anyone running into this issue is to simply run a `git mv` or try renaming the submodule so that their names and paths match.

vito:
fixes #246

@norbertbuchmueller Would you mind taking a look at this if you have time? Thanks if so!

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -136,12 +136,19 @@ if [ "$submodules" != "none" ]; then
 
     submodules="$(git config --list --name-only | sed -ne 's/^submodule\.\(.*\)\.url$/\1/p' | sort -u)"
   else
-    submodules=$(echo $submodules | jq -r '(.[])')
+    submodules=$(
+      echo $submodules | jq -r '.[]' | while read path; do
+        git config --file .gitmodules --name-only --get-regexp '\.path$' $path |
+          sed -e 's/^submodule\.\(.\+\)\.path$/\1/'
+      done
+    )
 
     git submodule init $submodules
   fi
 
   for submodule in $submodules; do
+    submodule_path=$(git config --file .gitmodules --get "submodule.${submodule}.path")
+
     if [ "$depth" -gt 0 ]; then
       # remember submodule update config
       update_conf_was_set=0
@@ -153,7 +160,7 @@ if [ "$submodules" != "none" ]; then
       git config "submodule.${submodule}.update" "!$bin_dir/deepen_shallow_clone_until_ref_is_found_then_check_out $depth"
     fi
 
-    git submodule update --no-fetch $depthflag $submodule_parameters $submodule
+    git submodule update --no-fetch $depthflag $submodule_parameters $submodule_path
 
     if [ "$depth" -gt 0 ]; then
       # restore submodule update config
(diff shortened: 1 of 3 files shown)


assets/in
```diff
@@ -136,12 +136,19 @@ if [ "$submodules" != "none" ]; then
...
   else
-    submodules=$(echo $submodules | jq -r '(.[])')
+    submodules=$(
+      echo $submodules | jq -r '.[]' | while read path; do
+        git config --file .gitmodules --name-only --get-regexp '\.path$' $path |
+          sed -e 's/^submodule\.\(.\+\)\.path$/\1/'
+      done
+    )
 
     git submodule init $submodules
```
norbertbuchmueller: According to the documentation, `git submodule init` also expects paths (and not names), so i think this is wrong as `$submodules` is the list of _names_ of the submodules here.

I think it would make things simpler and cleaner if we would work with submodule _paths_ everywhere, as that's what is passed in via the `submodule` parameter and that's what the `git submodule` sub-commands expect. Then we could avoid the extra steps to convert them to names and then converting the names back to paths. It would be then a single change to line 137 plus the test.

vito: I tried this for a long time but for reasons I could not understand it kept breaking the submodule deepening test. It seemed like the update hook that we set in the config wasn't being used at all. 

Speaking of which, I believe we need the name for configuring that hook.

rizwanreza:
Is there any timeline on when this will get merged? Thanks.

vito:
@rizwanreza I was pretty spooked about the simpler approach not working, but I guess I can just merge this and let y'all try it out. We can always try the simpler approach later.

PR #248, opened 2019-03-20
Bring LFS functionality up to date

vectorstorm:
In recent releases of git, LFS (Large File Storage) has been integrated into the git core, and the way that the git-resource has been dealing with LFS has become out of date.  Most notably, the `disable_git_lfs: true` parameter no longer works using the version of git included in current containers.

The major change to git has been that `git clone` will now automatically set up LFS and pull down LFS objects by default (when the repo uses LFS), without requiring any `git lfs` commands to be issued at all.

The first commit in the series removes the fetch of LFS objects from the 'check' script (as we don't need them when we're just looking for commits), and the second commit fixes the `disable_git_lfs` parameter to work again in the 'in' script;  when specified, we once again don't pull down LFS objects.  

Both changes are implemented by setting a `GIT_LFS_DISABLE_SMUDGE=1` environment variable to short-circuit the LFS process, as documented [here](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-config.5.ronn).

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -38,6 +38,9 @@ if [ -n "$tag_filter" ]; then
   tagflag="--tags"
 fi
 
+# We're just checking for commits; we don't ever need to fetch LFS files here!
+export GIT_LFS_SKIP_SMUDGE=1
+
 if [ -d $destination ]; then
   cd $destination
   git fetch $tagflag -f
--- a/assets/in
+++ b/assets/in
@@ -73,6 +73,11 @@ if [ -n "$tag_filter" ]; then
   tagflag="--tags"
 fi
 
+if [ "$disable_git_lfs" == "true" ]; then
+  # skip the fetching of LFS objects for all following git commands
+  export GIT_LFS_SKIP_SMUDGE=1
+fi
+
 git clone --single-branch $depthflag $uri $branchflag $destination $tagflag
 
 cd $destination
@@ -105,11 +110,6 @@ if [ ! -z "${commit_verification_keys}" ] || [ ! -z "${commit_verification_key_i
       xargs --no-run-if-empty -n1 gpg --batch --keyserver $gpg_keyserver --recv-keys
   fi
   git verify-commit $(git rev-list -n 1 $ref) || commit_not_signed
-fi
-
-if [ "$disable_git_lfs" != "true" ]; then
-  git lfs fetch
-  git lfs checkout
 fi
 
 git log -1 --oneline
@@ -164,10 +164,6 @@ if [ "$submodules" != "none" ]; then
       fi
     fi
   done
-fi
-
-if [ "$disable_git_lfs" != "true" ]; then
-  git submodule foreach "git lfs fetch && git lfs checkout"
 fi
 
 for branch in $fetch; do
--- a/test/all.sh
+++ b/test/all.sh
@@ -6,5 +6,6 @@ $(dirname $0)/image.sh
 $(dirname $0)/check.sh
 $(dirname $0)/get.sh
 $(dirname $0)/put.sh
+$(dirname $0)/lfs.sh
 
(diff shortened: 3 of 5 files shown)


vectorstorm:
This also resolves the following warnings visible in log output from 'get' jobs:

> WARNING: 'git lfs checkout' is deprecated and will be removed in v3.0.0.
'git checkout' has been updated in upstream Git to have comparable speeds	
to 'git lfs checkout'.

vectorstorm:
Found a bug in the original pull-request; the check script was still pulling LFS objects for files modified after the original clone.  Now we set the `GIT_LFS_DISABLE_SMUDGE=1` environment variable universally within the check script so that it affects both the initial clone and also subsequent fetches, instead of setting it only for the clone itself.

vito:
Followup question: do we still need to install git LFS separately?

vectorstorm:
Yes, we do still need to install git-lfs explicitly;  the tighter integration is achieved via git hooks installed globally by the git-lfs apk during its installation, so if it's not installed there's no change from vanilla git.  (specifically, the apk for git-lfs installs a file `/etc/gitconfig` which sets the required lfs filters to run automatically post-checkout and pre-commit)

**This is different from most distro's git lfs packaging.**  On Debian, for example, after installing the git-lfs package the user must then run 'git lfs install' explicitly in order to set up those filters and just for themselves, but alpine's git-lfs apk does that setup automatically and globally.  This is something to keep an eye on if this image ever gets ported to a different base.  (Maybe it would be worth us calling 'git lfs install' explicitly in the Dockerfile, just to guard against that?  It shouldn't cause any harm, and would ensure everything keeps working even if git-resource switches to a Ubuntu base or whatever)

Also -- would it be worth me adding a couple tests to verify the correct working of the disable_git_lfs parameter, and possibly of lfs itself?  I notice that nothing seems to be testing them right now.  I could put something together and add it as a third commit in this pull request.  Caveat:  Doing lfs tests would require network connectivity;  lfs doesn't work for cloning repos directly from the local filesystem.

vectorstorm:
...although, GitHub has a relatively small [git lfs server program](https://github.com/git-lfs/lfs-test-server/releases) which we could perhaps load into the container to test real lfs functionality locally.  That might be the way to go.

vectorstorm:
Okay, these changes are now functional.

As mentioned above, in addition to the previous changes in the first two commits, I've added a 'git lfs install' run statement to the Dockerfile, as a reminder to folks porting to Ubuntu or other images that it's required there.  (commit 3)

I've also added tests which verify that LFS is working, and also that the `disable_git_lfs` parameter is successfully disabling LFS functionality.  I've put those in a separate 'lfs.sh' script in the 'test' directory, so that they can easily be excluded if desired.  Note that the LFS tests are currently fetching a tiny repo owned by me to verify LFS functionality (the test server that I mentioned earlier turned out to be problematic for use from a script, so I went with github instead).  That repo should be forked by concourse (or one like it should be created from scratch), and the test modified to clone from the concourse-owned version of it before this pull request is actually merged.  (commit 4)

Apart from needing that concourse-owned LFS test repo, I think this is basically ready to go, now!

vectorstorm:
With the restructuring of the git-resource repo, the last set of commits weren't merging cleanly any more.  I've rebased changes onto the new master branch, and removed the "git lfs install" commit, since it wasn't needed in Alpine and the new Ubuntu-based Dockerfile is handling lfs already.

vito:
@vectorstorm Glad you noticed the `alpine` vs. `ubuntu` change, funny that you called out the packaging differences just as we were on the verge of switching to Ubuntu. :+1:

I'll merge this soon and migrate the test repo over. Thanks!

vito:
Merged - thanks!

Issue #198, opened 2018-06-19
Checkout non-detatched mode


Jarlotee:
It looks like currently git always checks out master in detached mode.

But I need to reference what branch has been selected for some automation scripts

Is there parameter to toggle the detached behavior?

Finkregh:
is this under consideration? This has been open for a year now and did not get any feedback :(

vito:
Seems like an easy feature to PR. The resource could write to `.git/branch` just like it writes to `.git/ref` already.

PR #257, opened 2019-04-23
Allow cloning into a non-detached state

miloszrolinski:
Hello!
This pull request is filed to fix issue concourse/git-resource#198 and allow cloning into a non-detached state. Also, as suggested by @vito I have added a single line of code to write the original branch into .git/branch if someone does not need a non-detached clone/symbolic HEAD but would like to have this information anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,8 @@ correct key is provided set in `git_crypt_key`.
 
 * `disable_git_lfs`: *Optional.* If `true`, will not fetch Git LFS files.
 
+* `no_detach_head`: *Optional.* If `true`, will not produce a repo in *detached HEAD* state, leaving HEAD as symbolic ref to branch, instead of a direct reference to a commit.
+
 * `clean_tags`: *Optional.* If `true` all incoming tags will be deleted. This
   is useful if you want to push tags, but have reasonable doubts that the tags
   cached with the resource are outdated. The default value is `false`.
--- a/README.md
+++ b/README.md
@@ -193,6 +193,8 @@ the case.
  * `.git/ref`: Version reference detected and checked out. It will usually contain
    the commit SHA-1 ref, but also the detected tag name when using `tag_filter`.
 
+ * `.git/branch`: Name of the original branch that was cloned.
+
  *  `.git/short_ref`: Short (first seven characters) of the `.git/ref`. Can be templated with `short_ref_format` parameter.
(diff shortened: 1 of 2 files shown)


vectorstorm:
What does this do in the case of a pinned resource?

vito:
Yeah - this seems to just leave it at whatever commit was on the branch at the time the clone happened, with no guarantee that the tree points to the version we're trying to fetch.

Maybe it should do a `git reset --hard $ref`?

vectorstorm:
I don't think that `git reset --hard $ref` works in some situations.

As one example:  Imagine that we've set a tag filter, and haven't specified any particular branch, so our resource is picking up every commit with a tag that matches the filter, regardless of what branch those commits are on.  When the 'get' step does the clone, if we haven't specified a specific branch then we're checking out the default branch (typically master), and then when we do the proposed `git reset --hard`, we make that default branch point to the matched tag commit, no matter where that tag actually came from.

This pull request's implementation of writing out the "current branch" to a file has the same problem;  it will only work if the optional "branch" parameter is specified explicitly;  otherwise, we'll claim that whatever commit we've gotten is coming from the default branch (typically master).  

(If we're okay with needing the optional "branch" parameter in order to provide the branch information file, then we could just echo out the bash variable `$branch` to the file, rather than calling into git+sed at all!)

A more robust solution might be to check out the specific commit as we do now, and then call `git branch -r --contains HEAD`, and do a little sed work to massage the output;  it provides a list of branches which contain the current commit;  take your pick as to which of them is the "correct" branch to claim represents the commit.

arcticShadow:
Hello - Has there been any further thinking on this issue? 

I've come on an issue hunt to see if there was any interesting issues around detached heads - and come across #198 which has lead me here. While i wont pretend to understand the code that powers this resource, nor the implications that are discussed, I would like to respectfully ask if there is a way that this issue can currently be worked around?

gerhard:
I have just come across a situation where having this option would have come in handy.

I worked around not having this by running `git checkout  "$GIT_BRANCH"` in the task, before adding & committing the changes, so that the `put` can do the right thing.

jamieklassen:
I'm going to close this now due to inactivity, but leave the discussion - if anybody involved wants to revisit this it would be OK to open a new PR. Just be sure to add tests!

PR #261, opened 2019-05-02
initial resource fetch should only use tags off of branch if specified

palixthepalalix:
We noticed after an upgrade to concourse 5.1.0, after unpausing all the pipelines, resources that were configured to only pull in tags off of master were getting tags that were on other branches. The resources looked something like this:
```
resources:
- name: code
  type: git
  source:
    branch: master
    private_key: ((PRIVATE_KEY))
    tag_filter: '*'
    uri: git@github.com:user/somerepo.git
```
I reproduced this issue by creating a new branch in a repo brand new repo and running the following steps:
```
cd somerepo
# on master, master has no tags
git checkout testbranch
touch file
git add file
git commit -m 'adding file' && git push origin HEAD
git tag -a 0.1.0 -m 'tagging 0.1.0' &&  git push origin 0.1.0
```
With a pipeline triggered on the resource above, you'll see `0.1.0` show up in the check in concourse, even though this tag is on `testbranch` and we're only looking for tags on master. 

This is related to the issue resolved in this PR https://github.com/concourse/git-resource/pull/260

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -103,7 +103,11 @@ if [ -n "$tag_filter" ]; then
       tags=$(git tag --list "$tag_filter" --sort=creatordate | lines_including_and_after $ref)
       get_commit $tags
     else
-      tag=$(git tag --list "$tag_filter" --sort=creatordate | tail -1)
+      branch_flag=
+      if [ -n "$branch" ]; then
+        branch_flag="--merged $branch"
+      fi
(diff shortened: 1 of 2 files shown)


palixthepalalix:
Those build failures don't seem to be related to my changes:
```waiting for docker to come up...
Error response from daemon: Get https://registry-1.docker.io/v2/: toomanyrequests: too many failed login attempts for username or IP address```

PR #262, opened 2019-05-21
Add "returning" option for merge output ref

### LINKED ISSUE #253: Use original commit version as output version after a merge
Currently, if the `merge` parameter is set for `out`, the version ref used will be the result of the merge commit. I believe the version ref should instead be the original commit.

Consider my poorly drawn diagram of a commit history:

```
*  m - Merge commit of x and o
|\
* |  x - Commit we're racing against
| *  o - Good commit that we want
|/
* p - Parent commit
```

And my poorly drawn pipeline:
```
git-repo --> [ run-tests ] --> git-repo --> [ bump-version-and-tag ] --> git-repo-bumped --> [ deploy ]
```

In this case, I've made some changes in commit `o` as part of my pipeline, and I'd like to tag and push commit `o`. To prevent pipeline failures if some commit `x` ends up in my target branch before the pipeline is finished, I've set `merge: true`, which merges the latest target branch here, combining `x` and `o` into merge commit `m`:
https://github.com/concourse/git-resource/blob/9129dd95addcc9ab69a1e4bd3f0d174de80b477c/assets/out#L177-L192

And then ultimately uses the current commit, which would be commit `m`. This means the rest of my pipeline will now include commit `x`, despite it not being part of the relevant changeset:

https://github.com/concourse/git-resource/blob/9129dd95addcc9ab69a1e4bd3f0d174de80b477c/assets/out#L199-L202

Because merging creates a non-linear history where commit `o` will still be intact, I think commit `o` should be used as the version ref. This would mean future parts of the pipeline can operate on the original change set without picking up racing commits in the target branch, while still being able to output the change into the target branch without races.

Edit: If this is something you're interested in, I'd be happy to contribute the change.

rliebz:
Setting the `returning` param to `unmerged` allows using the original commit as the resource output when setting the `merge` param to `true`, rather than using the merge commit. This is a backward compatible change that allows users to opt-in to the new behavior as desired.

Resolves #253

### DIFF
--- a/test/put.sh
+++ b/test/put.sh
@@ -274,6 +274,43 @@ it_can_put_to_url_with_merge_commit() {
   test $latest_merge_ref = $merged_ref
 }
 
+it_chooses_the_unmerged_commit_ref() {
+  local repo1=$(init_repo)
+
+  local src=$(mktemp -d $TMPDIR/put-src.XXXXXX)
+  local repo2=$src/repo
+  git clone $repo1 $repo2
+
+  # make a commit that will require rebasing
+  local baseref=$(make_commit_to_file $repo1 some-other-file)
+
+  local unmerged_ref=$(make_commit $repo2)
+
+  # cannot push to repo while it's checked out to a branch
+  git -C $repo1 checkout refs/heads/master
+
+  local response=$(mktemp $TMPDIR/rebased-response.XXXXXX)
+
+  put_uri_with_merge_returning_unmerged $repo1 $src repo > $response
+
+  local merged_ref=$(git -C $repo2 rev-parse HEAD)
+
+  jq -e "
+    .version == {ref: $(echo $unmerged_ref | jq -R .)}
+  " < $response
+
+  # switch back to master
+  git -C $repo1 checkout master
+
+  test -e $repo1/some-file
+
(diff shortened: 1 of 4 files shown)


test/put.sh
```diff
@@ -274,6 +274,43 @@ it_can_put_to_url_with_merge_commit() {
   test $latest_merge_ref = $merged_ref
 }
 
+it_chooses_the_unmerged_commit_ref() {
```
vito: You'll need a `run it_chooses_the_unmerged_commit_ref` line at the bottom of the file in order to run the test. :slightly_smiling_face:

rliebz:
@vito -- Looks like [the Ubuntu test fails sometimes](https://ci.concourse-ci.org/teams/resources/pipelines/git/jobs/prs-ubuntu/builds/28), so I had to re-trigger the build (with an amend/force-push) to get it to pass. There are other examples of the same error in the build history ([26](https://ci.concourse-ci.org/teams/resources/pipelines/git/jobs/prs-ubuntu/builds/26), [18](https://ci.concourse-ci.org/teams/resources/pipelines/git/jobs/prs-ubuntu/builds/18)), so I don't think it's related to this change.

Other than that, the PR should be good to go.

PR #271, opened 2019-09-25
Fix limiting output to 1 commit

CliffHoogervorst:
We have an issue with the current Git resource because it lists a lot of commits when looking at a new branch. 

This is because the location of  '-1' parameter in the `git rev-list` command is wrong:

`git rev-list --all --first-parent -- 'terraform/main/spaces/iam/*' -1` should be:
`git rev-list --all --first-parent -1 -- 'terraform/main/spaces/iam/*'`


CliffHoogervorst:
@vito I added a test for our usecase.

PR #273, opened 2019-10-03
fix reverse when using path

CliffHoogervorst:
Signed-off-by: Hoogervorst <cliff.hoogervorst@nn-group.com>

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -105,7 +105,7 @@ it_can_check_from_a_ref() {
   local ref1=$(make_commit $repo)
   local ref2=$(make_commit $repo)
   local ref3=$(make_commit $repo)
-
+  check_uri_from $repo $ref1
   check_uri_from $repo $ref1 | jq -e "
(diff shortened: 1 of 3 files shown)


test/check.sh
```diff
@@ -105,7 +105,7 @@ it_can_check_from_a_ref() {
   local ref1=$(make_commit $repo)
   local ref2=$(make_commit $repo)
   local ref3=$(make_commit $repo)
-
+  check_uri_from $repo $ref1
```
jomsie: I think this might've been added accidentally

jomsie:
Howdy!
Just ran the tests locally with this change, looks good! 

Just a few suggestions:
1. We have a WIP check, so the title will need to be changed to get rid of the WIP
2. Our DCO needs each commit to have an author sign-off
3. Could probably squash those commits

CliffHoogervorst:
Fix for issue #275

PR #277, opened 2019-10-04
Do not require branch if pushing only tags

alex-slynko:


grmaciel:
@alex-slynko would this allow to create/trigger tags from any branch? @vito any chance this could be reviewed?

alex-slynko:
@grmaciel Yes. It would allow creating tags from any ref.
In our case, we would be able to get commit ref from docker image, checkout repo to the given ref and then just do `put` for resource.

alex-slynko:
@YoussB Yes, that is correct. We get the git commit info from a different source and there is no need to specify the branch name at all. Anyway, it is not used in the resource itself.

YoussB:
@alex-slynko Happy to move it forward, but would appreciate adding a test around it.

alex-slynko:
@YoussB I have modified the existing test, so it does not need to send the branch.

PR #281, opened 2019-10-10
create `.git/describe_ref` to host tag

davidB:
This PR is a variant of PR #242.

- create the file `.git/describe_ref` from `git describe`
- allow to customize the git describe's options, so user can use unannoted tags,...


davidB:
any feedback ?

davidB:
Can someone review the PR ?

wagdav:
@davidB thanks for picking this up! I left a comment on #242 

@vito would you please take a look at this PR? Thanks!

README.md
```diff
@@ -183,6 +180,23 @@ correct key is provided set in `git_crypt_key`.
 
 * `short_ref_format`: *Optional.* When populating `.git/short_ref` use this `printf` format. Defaults to `%s`.
 
+* `describe_ref_options`: *Optional.* When populating `.git/describe_ref` use this options to call [`git describe`](https://git-scm.com/docs/git-describe). Defaults to `--always --dirty --broken`.
```
vito: I'm wary of exposing things like this because it adds a tight coupling to the `git` CLI and potentially exact versions of `git`. Can we remove it if it's not necessary? We usually avoid this kind of open-ended parameterization because it ties our hands later when we want to e.g. implement resources using SDKs and APIs instead of CLIs and having no clean way of interpreting these parameters backwards-compatibly.

davidB: different team could require different parametrization (previous PR doesn't select the same value as the default I propose), so (imho) it should not be a hardcoded configuration.

I thought it is clearer (for user, and for implementation) to use the same flags. Else it means recreate the exact mapping and documentation, parsing flags, and create the git call that will parse again. If the api of the future version of git included (into the resources) is not compatible with the current one, anyways it will be needed to write adaptation/translation or to break (if some features are removed). So I don't see the gain to encapsulate / to wrap today.

Anyway I'll search, if I found a way to be more aligned with your recommendation and to keep parametrization (without adding too many complexity/code/potential bug - easy in shell script).

PS: On another project I used libgit2 directly and trying to reproduce some "git cli" with lib is very complex and error prone. And as user of this resource, I expect it to use git cli.

tmduncan: Any movement here?  Would love to see this as a feature.

davidB: @vito will you prefer if I hide the cli behind a yaml map and then internally use jq to convert it into the cli ? something based on

```sh
~ ❯ echo '{"always": true, "dirty": true, "tags": false, "bar2": true}' | jq 'to_entries                                                                                                      
 | map(select(.key | test("^all$|^tags$|^long$|^first-parent$|^abbrev$|^exact-match$|^candidates$|^match$|^exclude$|^always$|^dirty$|^broken$")))
 | map(select(.value != false))
 | map(if .value == true then "--\(.key)" else "--\(.key) \(.value)" end)
 | join(" ")'

"--always --dirty"
```

- only some key are used (other are ignored)
- no check if the type of the value is correct (boolean or string) for the key
- It will be less powerfull (no support of `[=<n>]`) with this current code (maybe it could be done later if user request it or PR.

aledeganopix4d:
Hello, what is the status of this PR?

README.md
xtremerui: I agree that parametrization is needed for `git describe` as the command itself is for customization of human readable name. 

For coupling of `git` CLI I think this resource mean to using Git vcs. Since`git` is the official CLI, unless there is a plan that git-resource will switch over to an alternative CLI for Git I feel we can assume `git` CLI will be our LTS.

For coupling an exact versin of `git` CLI, I have checked [the changes of `git describe` since 2005](https://github.com/git/git/commits/master?before=20514004ddf1a3528de8933bc32f284e175e1012+35&path%5B%5D=Documentation&path%5B%5D=git-describe.txt), it looks like there isn't a single breakig change so far. Not future-proof but at least we can assume the maintenance in our end is trivial.

Combine all above @vito I would like to give green light to this PR (with the change request for below). WDYT?

README.md
```diff
@@ -183,6 +180,23 @@ correct key is provided set in `git_crypt_key`.
...
+    --long                always use long format
+    --first-parent        only follow first parent
+    --abbrev[=<n>]        use <n> digits to display SHA-1s
+    --exact-match         only output exact matches
+    --candidates <n>      consider <n> most recent tags (default: 10)
+    --match <pattern>     only consider tags matching <pattern>
+    --exclude <pattern>   do not consider tags matching <pattern>
+    --always              show abbreviated commit object as fallback
+    --dirty[=<mark>]      append <mark> on dirty working tree (default: "-dirty")
+    --broken[=<mark>]     append <mark> on broken working tree (default: "-broken")
```
xtremerui, line 197: To avoid the maintenance work in the doc could we remove this part, since [`git describe`](https://git-scm.com/docs/git-describe) is more clear and up-to-date already?

README.md
davidB: So keep the cli args or try to add some kind of convertion from yaml key-value map (used in config) to cli args (use internally). I ask to know if I should change this part of the PR.

xtremerui: IMO keeping the cli args is fine.

README.md
davidB: section removed from README, as requested

README.md
xtremerui: @YoussB do you have any comment here? If not I tend to merge this PR. Thx!

Issue #279, opened 2019-10-08
The common script should remove the prefix https://github.com


vinhtran2905:
From git resource log shows the URL incorrectly so that user can not click the git commit log 
 /https://github.abc.com/xxxxx

URL: https://github.com/https://github.abc.com/git-repo/postman/commit/62fccc3932fb51477eda97e1569e9c75060c99e5

--

add_git_metadata_url() {
  local commit=$(git rev-parse HEAD)
  local origin=$(git remote get-url --all origin) 2> /dev/null

  if echo $origin | grep github.com > /dev/null; then

    # git@github.com:concourse/git-resource.git     -> concourse/git-resource
    # https://github.com/concourse/git-resource.git -> concourse/git-resource
    local ownerRepo=$(echo $origin | sed -e' s/.*github.com[:\/]//; s/\.git$//')
    local url=$(echo "**https://github.com/**$ownerRepo/commit/$commit" | jq -R . )

    jq ". + [
        {name: \"url\", value: ${url}}
    ]"
  else
    jq ". + []"
  fi
}

agurney:
The example has `github.abc.com` but this actually affects just those URLs that have `github.com` anywhere in them, such as a GitHub Enterprise installation at `github.companyname.com`. The test could look for `(git@|https://)github.com[:/]` or something, in order to only catch genuine github.com URLs, and that would at least stop this particular example from triggering.

Aside from that I don't see a good general way to discover an "clickable commit URL" corresponding to a given hash and remote. I can _imagine_ actively probing something derived from the remote URL, in order to find out "aha, this is GitHub Enterprise, I'd better generate a clickable URL of the proper form". Or the resource gains a configuration knob for the URL pattern. But neither of these seem very attractive.

vinhtran2905:
It works perfectly. Thanks for the fix.

PR #287, opened 2019-11-05
Support more commit URL possibilities in metadata

agurney:
The existing logic in `add_git_metadata_url` tries to synthesize a URL of the form "https://github.com/org/repo/commit/abc123", in the case where we know we're working with a GitHub repo. The URL can be displayed in the Concourse UI, so users can click to navigate straight to the commit in question.

Issue #279 reported that this did not work correctly for a given GitHub Enterprise remote URL containing the substring "github.com", because the code just looked for that string, and tried to apply some straightforward rewrite rules in order to generate the commit URL.

More generally, it would be nice to be able to support these URLs for a wider range of hosting options.

In this change, we switch to parsing out various components of the remote URL, so that the target hostname can be identified more reliably; based on that hostname, we guess a commit URL format. The guessing logic should be enough to identify repos hosted on GitHub, GitLab or BitBucket shared hosting, plus self-hosted instances of any of them (or of gogs.io) in the case where the hostname includes "github", "gitlab", etc.

This therefore _doesn't_ identify a GitHub Enterprise instance living at something like "code.somebigcompany.com"; there's no way to tell this is GHE without either being told explicitly, or throwing traffic at it in order to guess. Neither of those options seem really attractive for the time being.

Additionally, there are a lot of ways that the source URI _might_ be set up. Only some of these are commonly used with the providers we're targeting here. This does support the following:

* https://github.com/org/repo.git
* ssh://git@github.com/org/repo.git
* ssh://git@github.com:443/org/repo.git (for if ssh is blocked)
* git@github.com:org/repo.git

and their non-GitHub parallels, which seem to represent the most typical configurations.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -135,15 +135,33 @@ add_git_metadata_url() {
   local commit=$(git rev-parse HEAD)
   local origin=$(git remote get-url --all origin) 2> /dev/null
 
-  if echo $origin | grep github.com > /dev/null; then
+  # This is not exhaustive for remote URL formats, but does cover the
+  # most common hosting scenarios for where a commit URL exists
+  if [[ ! $origin =~ ^(https?://|ssh://git@|git@)([^/]+)/(.*)$ ]]; then
+    jq ". + []"
+  fi
+  
+  local host=${BASH_REMATCH[2]}
+  local repo_path=${BASH_REMATCH[3]%.git}
 
-    # git@github.com:concourse/git-resource.git     -> concourse/git-resource
-    # https://github.com/concourse/git-resource.git -> concourse/git-resource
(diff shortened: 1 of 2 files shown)


agurney:
Whoops, I entirely missed the existing test suite. I'll fix those right up.

agurney:
the Ubuntu build failure (https://ci.concourse-ci.org/builds/7190740) comes from
```
error parsing HTTP 429 response body: invalid character 'T' looking for beginning of value: "Too Many Requests (HAP429).\n"
```

apparently from failing to push up the resource image once it is made? In any case, the 'actual' tests passed.

vito:
Hey thanks, good idea! I'll re-trigger the Ubuntu build, we've been hitting rate limits against Docker Hub lately and are working on a new environment so all our traffic isn't coming from one external IP.

PR #288, opened 2019-11-11
Add support for follow-flags

williammartin:
To support `yarn release` I'd like support for `--follow-tags`:

https://git-scm.com/docs/git-push#Documentation/git-push.txt---follow-tags

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -40,6 +40,7 @@ force=$(jq -r '.params.force // false' < $payload)
 only_tag=$(jq -r '.params.only_tag // false' < $payload)
 annotation_file=$(jq -r '.params.annotate // ""' < $payload)
 notes_file=$(jq -r '.params.notes // ""' < $payload)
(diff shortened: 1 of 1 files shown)


williammartin:
I meant to open this as a draft for discussion.

jamieklassen:
@williammartin I have converted to a draft (I don't think that feature existed when you first opened this PR) - are you still interested in this discussion? How do you feel about @YoussB's questions about the intended use-case for this change?

williammartin:
@jamieklassen I'm afraid I'm long past working on this as it was a hack at the time, and had already forgotten most of the context by the time it got picked up in April. 

My vague memory is that we had a `yarn release` step that created a tag and pushed a commit with that tag using this git flag.

I wouldn't be opposed to it being closed if that's easier, it might come around again at some point.

Thanks

jamieklassen:
@williammartin thanks for following up -- I will close this now, but if the need re-arises there can be a new PR.

PR #289, opened 2019-11-11
Support for using git-crypt with a gpg key

moredhel:
This PR adds support for using a GPG key to decrypt the git repository as an alternative to symmetric encryption.

This is currently being tested from the current docker image: `moredhel/git-resource`.

I will be testing this internally for the moment & will remove the WIP when I'm happy that it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,16 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
   unlock / decrypt the repository with `git-crypt`. To get the key simply
   execute `git-crypt export-key -- - | base64` in an encrypted repository.
 
+* `git_crypt_gpg_key`: *Optional.* No encoding
+  [git-crypt](https://github.com/AGWA/git-crypt) key. Setting this will
+  unlock / decrypt the repository with `git-crypt`.
+  The key should be stored in such a way that it can be directly imported
+  by doing the following `echo "${git_crypt_gpg_key} | gpg --import"`.
+
+  [git-crypt](https://github.com/AGWA/git-crypt) key. Setting this will
+  unlock / decrypt the repository with `git-crypt`. To get the key simply
+  execute `git-crypt export-key -- - | base64` in an encrypted repository.
+
 * `https_tunnel`: *Optional.* Information about an HTTPS proxy that will be used to tunnel SSH-based git commands over.
   Has the following sub-properties:
     * `proxy_host`: *Required.* The host name or IP of the proxy server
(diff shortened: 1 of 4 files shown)


xtremerui:
Hi, @moredhel , we turned this PR to draft as it looks like there is still work not done (the commented out code in dockerfile that caused concourse-ci/status-ubuntu failed).

Are you still interested at getting this PR through? Thx!

moredhel:
@xtremerui I am not working with concourse anymore.
I also can't access the concourse ci pipelines to see what is failing in the builds. If you want to pull it in then I will leave it to you & the team to do so. Otherwise feel free to close it.

xtremerui:
@moredhel thx for the response. If concourse team later decided to go with the PR we will make sure to credit you as original author. 

Could you explain a bit more on "you can't access concoures ci pipelines"? Do you mean you can't see the build log without login or with login by Github? To me `concourse-ci/status-alpine` is broken but `concourse-ci/status-ubuntu` works for me after login. Sry if that was the reason that blocked your debugging.

xtremerui:
Thx for the video clip. I think the problem happens when the box popped up for permission of download.

The login query created by clicking the login button contains a param `req` as following
```
Request URL: https://ci.concourse-ci.org/sky/issuer/auth/github?req=iyvxahlu7gxr6hfhsdbcfuufb
```
when the box popped up that intercept your action, it broke the pipe of login flow (dex server handles it but browser stops the redirect to github) and the next time you click login button, the same `req` is already expired here
https://github.com/dexidp/dex/blob/ebef257dcdade698763928d57387d88a6f87fcbb/server/handlers.go#L289-L295

I think if you use a different browser that doesn't have that pop out will fix the issue. Or you could try refresh the login page after you dismiss the pop out box, where it will generate a new `req` that allow you to perform the login flow again. I am not sure if the box would pop out though.

moredhel:
I was hopeful too earlier, but sadly I don't think that's the case as I tried chrome too

image

Yeah, it wasn't really a fair test.

xtremerui:
```
{"timestamp":"2020-04-16T15:32:35.101498896Z","level":"error","source":"atc","message":"atc.sky.callback.failed-to-issue-concourse-token","data":{"error":"user doesn't belong to any team","session":"7.7556"}}
```
uhh, right, our CI is configured to allow concourse team member on Github to login. So that's the log I found for your login that failed, which is totally fine but we need a better UI to handle the error message. We have an issue openned for it here https://github.com/concourse/concourse/issues/4572.

For contributor to view the build logs though, we used to expose the pr pipelines and build jobs but recently we migrated concoruse resource pipelines to the main team, which put those pipelines into private and we forgot to expose them again... we will do it now so you should be able to see the build log without login shortly. Sry!

Issue #270, opened 2019-08-27
Submodule with different credentials


dav9:
How can git-resource be configured to fetch a submodule that requires different credentials than the parent repo? Ideally the credentials should not be hardcoded in the submodule url.

PR #290, opened 2019-11-25
Add config for private submodules from external git servers

simeonkorchev:
This PR is related to the use case in which a given container repository hosted on a given git server has private git submodules, which are part from another git server.
The PR adds an additional optional `submodules` config section, which contains information regarding the git server that hosts the private git submodules.
The information includes the `host` of the git server, `username` and `password` for a git user, which has a permissions to read/write the contained submodule.
Related to: https://github.com/concourse/git-resource/issues/270

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,10 +52,10 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
   the `branch`. Patterns are [glob(7)](http://man7.org/linux/man-pages/man7/glob.7.html)
   compatible (as in, bash compatible).
 
-* `submodules`: *Optional.* List of credentials for HTTP(s) auth when pulling/pushing private git submodules which are not stored in the same git server as the container repository.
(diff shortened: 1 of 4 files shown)


simeonkorchev:
@vito  would you take a look at the PR?

README.md
```diff
@@ -52,6 +52,17 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
   the `branch`. Patterns are [glob(7)](http://man7.org/linux/man-pages/man7/glob.7.html)
   compatible (as in, bash compatible).
 
+* `submodules`: *Optional.* List of credentials for HTTP(s) auth when pulling/pushing private git submodules which are not stored in the same git server as the container repository.
```
vito: Can we call this `submodule_credentials` or something instead? I could see it being confused with the `submodules` param on the `get` step.

simeonkorchev: That's a good catch! I have corrected it in the latest commit

gowrisankar22:
@vito can you check and merge the PR. We are waiting for this feature

Issue #294, opened 2019-12-16
Use latest git version in the ubuntu dockerfile


YoussB:
# Background:
- Currently the ubuntu docker file is using the official git binary from `apt-get`
- This is using version `2.17.0` while the current git version is `2.24.1`.
- There is a `ppa` where we can install the latest version from, but still requires the installation of `software-properties-common` package in order to use `add-apt-repository` which might make inflate the image.

PR #295, opened 2019-12-16
build ubuntu image with the latest git version

YoussB:
- fixes: #294


vito:
I think in the long run we may want to be more conscious of which versions we pull in and maybe even track it as a dependency. :thinking: But this is good to at least get it trending forward to latest available for now.

YoussB:
@vito it makes a lot of sense to track the latest git. I will add an issue into that.

PR #300, opened 2020-01-16
Add config option to support cloning from codecommit via ssh

ebmeierj:
When using ssh to clone from codecommit you are required to specify the ID of the key you are using to connect via the `User` setting in the ssh config. This change allows that as a configuration option.

Reference: https://docs.aws.amazon.com/codecommit/latest/userguide/setting-up-ssh-unixes.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
       -----END RSA PRIVATE KEY-----
     ```
 
+* `private_key_user`: *Optional.* Enables setting User in the ssh config
+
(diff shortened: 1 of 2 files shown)


jamieklassen:
@ebmeierj are you still interested in this PR? I agree with @YoussB that it would be a handy feature, but also worth testing and your commit still requires a signoff.

ebmeierj:
I apologize, I've moved to another team and I'm no longer working with concourse. I'm not sure that I'll have time to address this with my current workload.

jamieklassen:
@ebmeierj thanks for your response - I'll close the PR for now, but if the need re-arises we can have a new one.

Issue #308, opened 2020-03-20
Error on add_git_metadata_message


heldersepu:
Getting error:
```/opt/resource/common.sh: line 131: /usr/bin/jq: Argument list too long``` 

https://github.com/concourse/git-resource/blob/f0d57ba2b7d43eb37dd001e14fe8bbb874b9a49a/assets/common.sh#L126-L132

the message is HUGE, I think that message should be truncated to prevent that error

PR #309, opened 2020-03-20
Truncate message at 10KB

heldersepu:
This fixes Issue #308

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -124,7 +124,7 @@ add_git_metadata_tags() {
 }
 
 add_git_metadata_message() {
-  local message=$(git log -1 --format=format:%B | jq -s -R .)
+  local message=$(git log -1 --format=format:%B | head -c 10240 | jq -s -R .)
 
   jq ". + [
     {name: \"message\", value: ${message}, type: \"message\"}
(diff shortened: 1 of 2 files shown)


heldersepu:
> This is a good catch, though I have never encountered such a huge commit message in my life!

Ye, first-time I've seen it too! Seem Github Enterprise when merging branches the default message is a concatenation of all commits (user still has the option to overwrite)...

looking into that test

heldersepu:
@YoussB Can you check why is the CI failing ...

aoldershaw:
@heldersepu looks like this line is failing on alpine: https://github.com/concourse/git-resource/pull/309/files#diff-daaa154e22be44798ffcddf12a70e03eR121

```sh
local message=$(shuf -zer -n20000  {A..Z})
```

You should be able to achieve something similar with:

```sh
local message=$(cat /dev/urandom | tr -dc A-Z | head -c 20000 ; echo '')
```

Source: https://unix.stackexchange.com/a/230676

heldersepu:
> @heldersepu looks like this line is failing on alpine: https://github.com/concourse/git-resource/pull/309/files#diff-daaa154e22be44798ffcddf12a70e03eR121
> 
> ```shell
> local message=$(shuf -zer -n20000  {A..Z})
> ```
> 
> You should be able to achieve something similar with:
> 
> ```shell
> local message=$(cat /dev/urandom | tr -dc A-Z | head -c 20000 ; echo '')
> ```
> 
> Source: https://unix.stackexchange.com/a/230676

@aoldershaw Done, I think we are good to go...

PR #310, opened 2020-03-24
Fix #293 wrong last commit reported

### LINKED ISSUE #293: wrong last commit reported with list_command | git rev-list --stdin
On a customer git it seems they had a weird commit with year 2106 instead of 2016 :

```
commit d1f0d5f316ffd6a7ba9e03198e4aae55b3d2efb1
Date:   Tue Dec 10 12:34:33 2019 +0100

...

commit 06a8e882889e6e164018271f000330d4c23004bd
Date:   Sun Feb 7 06:29:14 2106 -9039932

commit 86ddfd204e0f7b07492e220c9da1dfd0b422def2
Date:   Wed Jun 1 11:41:20 2016 +0200
```

Since https://github.com/concourse/git-resource/commit/c7ce409128070a3d4b7abd2643565851a1a02dc9 / https://github.com/concourse/git-resource/pull/271

`06a8e882889e6e164018271f000330d4c23004bd` is detected as the latest commit each time.

In this case the cmd is
```
git rev-list --all --first-parent   | git rev-list --stdin --grep="\[ci\sskip\]" --grep="\[skip\sci\]" --invert-grep --first-parent --no-walk| git rev-list --stdin -1
```

What is the purpose of this ?
Should it be reverted ? Or changed by `list_command+="| git rev-list --stdin --date-order -1"` ? Or simply `list_command+="| head -n1"` ?

talset:
Last commit reported could be wrong after a rebase or wrong commiter date.

Use `--date-order` to show commits in the commit timestamp order.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -81,19 +81,21 @@ fi
 list_command="git rev-list --all --first-parent $log_range $paths_search"
 if [ `echo $filter_whitelist | jq -r '. | length'` -gt 0 ]
 then
+    list_command+=" | git rev-list --stdin --date-order --first-parent --no-walk "
     whitelist_items=$(echo $filter_whitelist | jq -r -c '.[]')
     for wli in "$whitelist_items"
     do
-        list_command+=" | git rev-list --stdin --date-order --grep=\"$wli\" --first-parent --no-walk"
+        list_command+=" --grep=\"$wli\""
     done
 fi
 
 if [ `echo $filter_blacklist | jq -r '. | length'` -gt 0 ]
 then
+    list_command+=" | git rev-list --stdin --date-order --invert-grep --first-parent --no-walk "
     blacklist_items=$(echo $filter_blacklist | jq -r -c '.[]')
     for bli in "$blacklist_items"
     do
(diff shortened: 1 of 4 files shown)


vito:
Hey thanks for picking this up!

Do you think you could add a test for this? There was a test added in https://github.com/concourse/git-resource/pull/271 so it's good that that test still passes, but there should probably be another one added that covers the case y'all hit, just to make sure we don't break it again.

Thanks!

talset:
I finally choose to use `--date-order` with `--no-walk=unsorted`.

A dedicated test have been added and a `make_commit_to_future` is also used in different tests to validate the commits order.

I also did a small improvement on include/exclude construction cmd to avoid too many pipes.

I builded an image from this PR `cycloid/concourse-git-resource` @RichardBradley @simonjohansson don't hesitate to test it. (sample of pipeline to test it: https://paste-owl.cycloid.io/?ebcc1fa047e2cd55#HUvtkckLFksmWtN1VpbCsRLatWYuxvUZVmxdFvh3DGdG)

Issue #312, opened 2020-03-26
Constantly showing different tags sometimes 


Bluesboy:
I have task that pulls repo and checking tags there, however sometimes it doesn't shows some tags and thus I can't be sure that I have proper version. Here is resource
```
resources:
  - name: source-code
    type: git 
    icon: github-circle
    webhook_token: ((webhook-token))
    check_every: 4h
    source:
      uri: ((git.ssh-remote))
      private_key: ((git.private-key))
      branch: master
jobs:
  - name: build-pr-image
    serial: true
    plan:
      - get: source-code
        trigger: false

      - task: semver
        config:
          platform: linux
          image_resource:
            type: registry-image
            source:
              repository: bluesboy/semver
              tag: 3.0.0
          inputs:
          - name: source-code
          outputs:
          - name: version
          run:
            path: bash
            args:
            - -exc
            - |
              export VERSION=$(semver bump patch $(git -C source-code describe --tags --abbrev=0))-c$(git -C source-code rev-parse --short HEAD)
              echo ${VERSION} > version/version
```
I'm using https://github.com/fsaintjacques/semver-tool packed in Docker Image, to bump version (but I'm sure that is irrelevant)
It looks like that sometimes resource doesn't pull latest resource version

talset:
The resource actually clone (or fetch) the repo without `--tags flags`. It only get all tags when using tag_filter.

https://github.com/concourse/git-resource/blob/master/assets/check#L39-L59

If you want tags information without using tag_filter, we might need to implement a new parameter to force tags fetch.

talset:
Could you try the `fetch_tags: true` option ? (https://github.com/concourse/git-resource#parameters)

I just builded a new images from git-resource master

```
resource_types:
  - name: git
    type: docker-image
    source:
      repository: cycloid/concourse-git-resource
      tag: latest
```

Let us know if solve your issue

YoussB:
@talset @Bluesboy seems that a fix was merged, if I got that correctly?

talset:
> @talset @Bluesboy seems that a fix was merged, if I got that correctly?

Yes the PR have been merged (https://github.com/concourse/git-resource/pull/314) and should fix this issue

YoussB:
ok, thanks for confirming @talset.
I will close this issue but @Bluesboy or @talset feel free to reopen if needed 😁

Bluesboy:
@talset Thank you very much. Haven't noticed this issue for a while now.

PR #314, opened 2020-04-14
Add fetch_tags param fix #249 and #312

### LINKED ISSUE #249: Feature Request: Add --no-tags parameter for git clone
when concourse git resource tries to "git clone --single-branch", it also fetches tag objects which are not needed at all. 
It also cased git clone failed with the following message
fatal: did not receive expected object 6b00a8c61ed379d5f925a72c1987c9c52129d364
fatal: index-pack failed

Actually object 6b00a8c61ed379d5f925a72c1987c9c52129d364 is an tag object.
I have set the depth to 1, but git clone still shows the same error.

Thanks very much!

talset:
Add fetch_tags param in check and in.

If not defined, keep the previous behavior.
If true, get all the tags with --tags.
If false, don't fetch tags with --no-tags.

It should solve #249 (need to clone repository without tags fetched)
and #312 (need to fetch tags when not using tag_filter)

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
   the `branch`. Patterns are [glob(7)](http://man7.org/linux/man-pages/man7/glob.7.html)
   compatible (as in, bash compatible).
 
+* `fetch_tags`: *Optional.* If `true` the flag `--tags` will be used to fetch
+  all tags in the repository. If `false` no tags will be fetched.
+
 * `submodule_credentials`: *Optional.* List of credentials for HTTP(s) auth when pulling/pushing private git submodules which are not stored in the same git server as the container repository.
     Example:
     ```
@@ -98,7 +101,7 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
     * `proxy_user`: *Optional.* If the proxy requires authentication, use this username
   *   `proxy_password`: *Optional.* If the proxy requires authenticate,
       use this password
-    
+
 * `commit_filter`: *Optional.* Object containing commit message filters
   * `commit_filter.exclude`: *Optional.* Array containing strings that should
     cause a commit to be skipped
@@ -197,6 +200,11 @@ correct key is provided set in `git_crypt_key`.
   not pass a `paths` filter test from skewing the cloned history away from
   `version.ref`, this resource will automatically deepen the clone until
   `version.ref` is found again.
+
+* `fetch_tags`: *Optional.* If `true` the flag `--tags` will be used to fetch
+  all tags in the repository. If `false` no tags will be fetched.
+
+  Will override `fetch_tags` source configuration if defined.
 
 * `submodules`: *Optional.* If `none`, submodules will not be
   fetched. If specified as a list of paths, only the given paths will be
--- a/assets/check
+++ b/assets/check
@@ -25,7 +25,6 @@ branch=$(jq -r '.source.branch // ""' < $payload)
 paths="$(jq -r '(.source.paths // ["."])[]' < $payload)" # those "'s are important
 ignore_paths="$(jq -r '":!" + (.source.ignore_paths // [])[]' < $payload)" # these ones too
 tag_filter=$(jq -r '.source.tag_filter // ""' < $payload)
-fetch_tags=$(jq -r '.source.fetch_tags // ""' < $payload)
 git_config_payload=$(jq -r '.source.git_config // []' < $payload)
 ref=$(jq -r '.version.ref // ""' < $payload)
 skip_ci_disabled=$(jq -r '.source.disable_ci_skip // false' < $payload)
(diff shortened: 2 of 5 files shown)


aoldershaw:
Also, though running the `in` script alone works well (i.e. can clone a repo with or without tags), I'm struggling to produce the results I'm expecting when running in a pipeline.

Here's the pipeline definition I'm using:

```yaml
resource_types:
- name: updated-git
  type: registry-image
  source: {repository: concourse/git-resource, tag: pr-ubuntu-314}
resources:
- name: repo
  type: updated-git
  source:
    uri: https://github.com/concourse/git-resource.git
    branch: master
    fetch_tags: false
jobs:
- name: tags
  plan:
  - get: repo
    trigger: true
  - task: tags
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: ubuntu}
      inputs:
      - name: repo
      params:
        TERM: ansi
      run:
        path: bash
        args:
        - -c
        - |
            apt-get update && apt-get install -y git
            git -C repo tag
```

I would expect `git -C repo tag` to list no tags since I'm setting `fetch_tags` to `false`, but this doesn't seem to be the case (it is listing several tags). In contrast, if I run `git clone https://github.com/concourse/git-resource.git --no-tags` locally, and then run `git tag`, it lists no tags as expected.

Please let me know if you have any ideas why this is.

README.md
```diff
@@ -198,6 +201,11 @@ correct key is provided set in `git_crypt_key`.
   `version.ref`, this resource will automatically deepen the clone until
   `version.ref` is found again.
 
+* `fetch_tags`: *Optional.* If `true` the flag `--tags` will be use to fetches
+  all tags in the repository. If `false` no tags will be fetched.
+
+  Will override `fetch_tags` source configuration if defined.
+
```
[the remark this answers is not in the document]
talset: Having it in source params allow us to define a "default" behavior and override it on some get.

For example the same git resource need to trigger 3 jobs, 2 don't need tags but the 1 use it.

assets/check
```diff
@@ -37,8 +38,10 @@ configure_git_global "${git_config_payload}"
 destination=$TMPDIR/git-resource-repo-cache
 
 tagflag=""
-if [ -n "$tag_filter" ]; then
+if [ -n "$tag_filter" ] || [ "$fetch_tags" == "true" ] ; then
   tagflag="--tags"
+elif [ "$fetch_tags" == "false" ] ; then
+  tagflag="--no-tags"
```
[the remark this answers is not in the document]
talset: You are right, it make sense

talset:
Thanks for your review @aoldershaw 

I think you hit a different "issue" with the git-resource that might be impacting other params.

Could you try to set the fetch_tags value with quotes ?

```
resources:
- name: repo
  type: updated-git
  source:
    uri: https://github.com/concourse/git-resource.git
    branch: master
    fetch_tags: "false"
```

talset:
If it is fine for you @aoldershaw , I did the following commit that should fix this behavior

https://github.com/concourse/git-resource/pull/314/commits/b1d706c80e76ce5448d8daf5575f7571115402ee

assets/check
```diff
@@ -37,7 +38,9 @@ configure_git_global "${git_config_payload}"
 destination=$TMPDIR/git-resource-repo-cache
 
 tagflag=""
-if [ -n "$tag_filter" ]; then
+if [ "$fetch_tags" == "false" ] ; then
+  tagflag="--no-tags"
+elif [ -n "$tag_filter" ] || [ "$fetch_tags" == "true" ] ; then
   tagflag="--tags"
```
aoldershaw, line 44: Hmm...if `source.fetch_tags` is intended to set the default behaviour, I'm realizing now that setting `source.fetch_tags` to `false` and defining `source.tag_filter` may not be such an unusual case (whenever you clone the repo in a `get` step, you don't care about the tags, but you want to trigger jobs for only commits that match some `tag_filter`).

With the logic the way it is, however, this won't actually work - since `$fetch_tags` is `"false"`, we always clone the repo with `--no-tags`. When we later try to run `git tag --list "$tag_filter"`, we'll always end up with an empty list, meaning the resource won't produce any versions.

I wonder if we even need to consider `source.fetch_tags` in the `check` script...we only ever need to fetch tags when `tag_filter` is set. Perhaps this could become the following:

```suggestion
if [ -n "$tag_filter" ]; then
  tagflag="--tags"
else
  tagflag="--no-tags"
```

Note that this would only be for `check` (i.e. not `in`).

Does that make sense? Let me know if what I'm saying won't work.

talset: I had a look on `check` and I did some test with `--no-tag`. Tags are needed only when using `tag_filter`.

As you said, we should be able to remove `fetch_tags` from `check` and set `--no-tags` as default behavior.

I will do more tests next week before pushing the changes.

aoldershaw: Awesome, and thanks for being so responsive!

talset: I did more tests on the `check` no issue discovered. I think we are fine ;)

PR #316, opened 2020-04-15
Fix issue #315 - correctly deepen clones

### LINKED ISSUE #315: Deepen shallow clone does not deep correctly for depth > 1
https://github.com/concourse/git-resource/blob/f0d57ba2b7d43eb37dd001e14fe8bbb874b9a49a/assets/deepen_shallow_clone_until_ref_is_found_then_check_out#L47

I've encountered a case where the git-resource is not correctly deepening a clone. I've specified `depth: 2` and expect to get `version.ref` + its previous commit fetched. However, in the following edge-case the previous commit is not fetched correctly.

The script retries fetching with ever larger `--depth` until the desired ref is found. However, this depth is always relative to the latest HEAD of the remote branch. This may move forward while the retry loop is in progress. 

The correct behavior would be to use `git fetch --deepen` instead, see https://git-scm.com/docs/git-fetch#_options

JohannesRudolph:
This PR fixes #315

### DIFF
--- a/assets/deepen_shallow_clone_until_ref_is_found_then_check_out
+++ b/assets/deepen_shallow_clone_until_ref_is_found_then_check_out
@@ -10,6 +10,7 @@ readonly ref="$2"
 readonly tagflag="$3"
 
 declare total_depth="$depth"
+declare depth_from_ref=$((depth - 1))
 
 # A shallow clone may not contain the Git commit $ref:
 # 1. The depth of the shallow clone is measured backwards from the latest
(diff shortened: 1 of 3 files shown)


JohannesRudolph:
I've found out that bitbucket does not support `git fetch --deepen` because they're stuck in 2016... https://jira.atlassian.com/browse/BCLOUD-14418

assets/deepen_shallow_clone_until_ref_is_found_then_check_out
```diff
@@ -24,27 +27,28 @@ readonly tagflag="$3"
 git_dir="$(git rev-parse --git-dir)"
 readonly git_dir
 
-while ! git checkout -q "$ref" &>/dev/null; do
+while ! git checkout -q "$ref~$depth_from_ref" &>/dev/null; do
```
xtremerui, line 30: Why do we start from `$ref~$depth_from_ref` instead of `$ref` here? Thx.

JohannesRudolph: The idea is described in the comment block L15-25. Basically by specifying the depth parameter the user says: give me a shallow repo at ref `x` where both `git checkout x` and `git checkout x~depth_from_ref` will work. Now the problem is that the cloned repo is actually cloned at ref `y` which may be any number of commits behind `x`. It's redundant to deepen until `git checkout x` works, if we deepen until `git checkout x~depth_from_ref` works it's guaranteed that `git checkout x` works too.

PR #318, opened 2020-04-27
Add option to ignore branch for the tag filter

sfunkhouser:
We have a use-case where we want different pipeline behavior for master-tags vs non-master tags. This option allows the resource to specify an `ignore_branch` to be used in conjunction with a tag_filter to get all tags *not* on master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
 * `branch`: The branch to track. This is *optional* if the resource is
    only used in `get` steps; however, it is *required* when used in a `put` step. If unset for `get`, the repository's default branch is used; usually `master` but [could be different](https://help.github.com/articles/setting-the-default-branch/).
 
+* `ignore_branch`: The branch to ignore, to be used in conjunction with `tag_filter`. This cannot be used with `branch` filter, `branch` filter will take precedence.
 
 * `private_key`: *Optional.* Private key to use when pulling/pushing.
     Example:
@@ -98,7 +99,7 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
     * `proxy_user`: *Optional.* If the proxy requires authentication, use this username
   *   `proxy_password`: *Optional.* If the proxy requires authenticate,
       use this password
-    
+
 * `commit_filter`: *Optional.* Object containing commit message filters
   * `commit_filter.exclude`: *Optional.* Array containing strings that should
     cause a commit to be skipped
--- a/assets/check
+++ b/assets/check
@@ -22,6 +22,7 @@ configure_credentials $payload
 
 uri=$(jq -r '.source.uri // ""' < $payload)
 branch=$(jq -r '.source.branch // ""' < $payload)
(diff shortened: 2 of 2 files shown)


muntac:
@sfunkhouser Thanks for the PR! Are you running into a problem where the "tag_filter" is obtaining tags from multiple branches instead of just the default branch when the `branch` field is not specified? I believe this is a bug, and was mentioned in this [comment](https://github.com/concourse/git-resource/pull/257#pullrequestreview-402869744) (jump to "tag_filter without branch").

It would be better to ensure `tag_filter` only pulls from the default branch (when branch is not specified) rather than allow users to specify specific branches to ignore.

sfunkhouser:
@muntac No, actually we want the tag_filter to pull from _all_ branches when its not specified. This is for a case where we want tags for the master branch to be treated differently than tags from non-master branches.

muntac:
@sfunkhouser Sorry about the delayed response! As @jamieklassen noted in in the [PR](https://github.com/concourse/git-resource/pull/257#pullrequestreview-402869744) I linked, the expected behaviour of the `git-resource` is to track a single branch thereby ignoring all other branches by default. The fact in some cases it doesn't do that is a bug.

In the [README](https://github.com/concourse/git-resource/blob/master/README.md) it's noted that when the branch is not set the resource will point to the repository's default branch, so we always have the notion of a single branch.

> branch: The branch to track. This is optional if the resource is only used in get steps; however, it is required when used in a put step. If unset for get, the repository's default branch is used; usually master but could be different.

Similarly, the `tag_filter` is meant to detect commits from a single branch. 

> tag_filter: Optional. If specified, the resource will only detect commits that have a tag matching the expression that have been made against the branch. Patterns are glob(7) compatible (as in, bash compatible).

I haven't been able to figure out exactly why the "tag_filter" is pulling from multiple branches when no branch is specified. I suspect this [line](https://github.com/concourse/git-resource/blob/03d1d2b099f605eaf2b0a93594632fb2a5543d75/assets/check#L142) needs to have a `--contains` or `--merged` like in the case when branch and ref is specified.

Either way, the current behaviour is a bug that needs to be fixed. Adding the ability to ignore some branches and pull from several others would rely on the bug. Sorry if this unintentional behaviour is something you've come to depend on!

sfunkhouser:
@muntac Good to know. We actually saw the docs and told our users internally for a long time that it should default to `master` if `branch` isn't specified. But then someone tested it and it worked so many teams started using it to be able to do builds on all tags, no matter the branch. 

Is there an open ticket that specifies this as a bug? We'd like to follow that to be able to implement a solution on our side that uses this correctly ahead of a bug fix.

sfunkhouser:
@muntac I'm going to close this PR since as you mentioned this is taking advantage of a bug in the resource. But if you could let me know if there is an open issue tracking the underlying issue that would be great. As mentioned in the linked PR: 

```
This is not what the documentation indicates, but I find it
kinda cool, and it's possible that users like @vectorstorm are depending on it.
```

We are definitely depending on it throughout many pipelines and want to be sure things do not break if/when this is fixed.

PR #322, opened 2020-05-11
copy netrc file to out dir to consume via load_var step

gowrisankar22:


gowrisankar22:
@pivotal-jamie-klassen Can you have look at this PR?

gowrisankar22:
@pivotal-jamie-klassen It makes sense but still netrc is stored as plaintext local to the get step container. This also, in my opinion, a security problem. We can have an encrypted netrc file and should be decoded during the runtime so that we are safe either case.

I also made a change from your proposal that, by default, this feature is switched off until unless somebody explicitly enables it a params `params: {output_netrc: true}`. 

It would be great if you can think of encrypting the netrc file itself.

vito:
Sorry but I really don't think this is a good idea. 😕 Resources should be careful with their credentials and not pass them around, and I don't think making it parameterizable really makes it any more OK. This feels like a workaround, or glue code designed for a very specific workflow - it shouldn't be in the core set of resource types, which are meant to be exemplary and idiomatic.

gowrisankar22:
@vito I was blocked from the discord channel. I am not sure what was the reason for blocking. Please let me do and don’ts so that it will help me to follow.

Sorry for the inconvenience and please do unblock my user..

vito:
@gowrisankar22 I gave a reason in the Discord UI but I guess it doesn't show it to the recipient. You were banned for repeatedly direct messaging server members after being asked several times not to. Discord isn't for 1:1 support.

This rule doesn't seem to be in the `#welcome` message at the moment, so I'll lift the ban for now, but please be more respectful of other peoples' time and focus from here on. One on one synchronous support sessions do not scale.

gowrisankar22:
@vito sure.sorry once again.  Still I am unable to join back ..

vito:
Sorry, got pulled away - should be unbanned now.

PR #327, opened 2020-07-30
Support off branch commits

TomWoodward:
I'm not 100% sure this change is desired, if there is interest i can add some more tests.

I have a config store of git refs for various repos i want to use in my build, I ran into two problems trying to implement that flow that i've fixed in this PR.

1) if my commits weren't on master (which is a little bit the point of having the decoupled config) this resource wouldn't be able to find them because of the way `git clone --single-branch` works.
1) initially i tried to use `load_var` with the default version pinning mechanic via `version` to get the ref i wanted. I'm actually not even sure if the two can be used together but regardless i found that the pinning would fail if the specified version was not already known to concourse

so to make my use case work i've added a `checkout` param to `put` that you can use to inform concourse of versions it didn't know about already. the resulting resource will be at the given version so no pinning is necessary you can just use `passed` as usual and get the version you want downstream.

i've changed `git clone --single-branch` to `git init && git remote add...`. my use case with the random commit reference is actually not enabled by default for git servers, but it works with github and that is good enough for me. if the server rejects the request for the specific commit i have it fall back on fetching the branch and i think that'll basically do the same as it was doing before.

when i got into checking the tests i realized that one of them was actually doing exactly what i wanted (it_omits_empty_branch_in_metadata fetches an unadvertised commit on a different branch). i questioned my sanity for about 2 days of debugging before realizing that the tests are using `git clone /path/to/other/repo` instead of `git clone file:///path/to/other/repo` and the former doesn't use the same git protocol and includes more data than you would expect, making the test work. i went ahead and changed all the clone paths and that one test now uses my new `in` logic.

### DIFF
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -96,8 +96,8 @@ add_git_metadata_committer() {
 }
 
 add_git_metadata_branch() {
-  local branch=$(git show-ref --heads | \
-    sed -n "s/^$(git rev-parse HEAD) refs\/heads\/\(.*\)/\1/p" |  \
+  local branch=$(git ls-remote origin | \
+    sed -n "s/^$(git rev-parse HEAD)\trefs\/heads\/\(.*\)/\1/p" |  \
     jq -R  ". | select(. != \"\")" | jq -r -s "map(.) | join (\",\")")
 
   if [ -n "${branch}" ]; then
--- a/assets/in
+++ b/assets/in
@@ -34,9 +34,9 @@ configure_git_ssl_verification $payload
 configure_credentials $payload
 
 uri=$(jq -r '.source.uri // ""' < $payload)
-branch=$(jq -r '.source.branch // ""' < $payload)
+branch=$(jq -r '.source.branch // "HEAD"' < $payload) # HEAD will reference the default branch on the remote
 git_config_payload=$(jq -r '.source.git_config // []' < $payload)
-ref=$(jq -r '.version.ref // "HEAD"' < $payload)
+ref=$(jq -r '.version.ref // ""' < $payload)
 depth=$(jq -r '(.params.depth // 0)' < $payload)
 fetch=$(jq -r '(.params.fetch // [])[]' < $payload)
 submodules=$(jq -r '(.params.submodules // "all")' < $payload)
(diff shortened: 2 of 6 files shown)


TomWoodward:
I found a variety of issues in the tests, i'm open to direction on better ways to handle those.

xtremerui:
@YoussB I think the first question is like what @TomWoodward mentioned:

Should it find a commit from a different branch?

xtremerui:
@TomWoodward What is the difference between `git clone /path/to/repo` and `git clone file://path/to/repo` ? Is the difference at `.git/objects` folder according to git documentation?

If that is the case, seems `git clone /path/to/repo --no-local` also does the same, would `--no-local` introduce less changes here?

TomWoodward:
@xtremerui the [documentation](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/git-clone.html) says:
> If the repository is specified as a local path (e.g., /path/to/repo), this is the default, and --local is essentially a no-op. If the repository is specified as a URL, then this flag is ignored (and we never use the local optimizations). Specifying --no-local will override the default when /path/to/repo is given, using the regular Git transport instead.

but that would mean adding --no-local to the resource itself, instead of limiting the handling of this test configuration to the tests, which seems a little iffy to me.

i can make that change if you want, but i'd like to hear back about whether the overall change is desired or not. if not we can make a separate pr to normalize the testing behavior

xtremerui:
@YoussB Could you give some inputs here?

xtremerui:
@TomWoodward i will close this due to the feature is implemented in https://github.com/concourse/git-resource/pull/333. If there is any concerns feel free to reopen.

TomWoodward:
@xtremerui pretty sure that change doesn't handle my use case of arbitrary commits on an unknown branch.

my use case is that _usually_ the tip of master is built, but on occasion we might manually lock in a different commit that might not even be on master (like a branched hotfix) 

i don't want to id release assets with a branch name because it could change under us without realizing

i could tag my hotfixes, but then i'd have inconsistent refs between hotfixes and normal continuous releases (tags vs commits)

the only way i could use the existing git-resource i think would be to auto-tag every commit on master, which i'm not going to do.

I don't feel like my needs are _super_ niche, but i'm ok running off of a fork if nobody else cares enough

PR #166, opened 2018-01-17
Allow selecting branches and tags via Perl regex

tprobinson:
There's a few other passenger commits in here, which I'd be happy to remove if you don't want them:

**dumb-init**: This can help reduce container process buildup issues by properly reaping children. Likely only an issue if a process has an error and doesn't clean up after itself, but dumb-init is small and has very little overhead.

**proxy vars**: I was building and testing this on my Concourse in a proxied environment, so I tried to adapt it to build in a proxied environment. Changing wget -> curl worked for the basic build, but the tests would still fail. I couldn't seem to get GPG to respect the proxy variables, it seemed like somewhere along the way the environment was getting reset.

**Skip tests**: The proxy issues are why I included the ability to skip the tests.

PR #339, opened 2020-11-13
Support filtering tags by regular expression

### LINKED ISSUE #232: tagfilter - prefer regex
While glob is decent, regex allows for more precise matching patterns.

jlsalmon:
This patch adds a new `tag_regex` option, which supports filtering tags by an egrep-compatible regex.

There was a previous attempt to implement this using Perl regexes (#166) but it didn't end up getting merged. I've opted to simply use grep here, as It seemed like a simpler approach while achieving the same outcome.

Fixes #232

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -570,6 +570,24 @@ it_can_check_with_tag_filter() {
   "
 }
 
+it_can_check_with_tag_regex() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  local ref2=$(make_annotated_tag $repo "1.0-staging" "tag 1")
+  local ref3=$(make_commit $repo)
+  local ref4=$(make_annotated_tag $repo "1.0-production" "tag 2")
+  local ref5=$(make_annotated_tag $repo "2.0-staging" "tag 3")
+  local ref6=$(make_commit $repo)
+  local ref7=$(make_annotated_tag $repo "2.0-staging" "tag 5")
+  local ref8=$(make_commit $repo)
+  local ref9=$(make_annotated_tag $repo "2.0-production" "tag 4")
+  local ref10=$(make_commit $repo)
+
+  check_uri_with_tag_regex $repo ".*-staging" | jq -e "
+    . == [{ref: \"2.0-staging\", commit: \"$ref6\"}]
+  "
+}
+
 it_can_check_with_tag_filter_with_cursor() {
   local repo=$(init_repo)
   local ref1=$(make_commit $repo)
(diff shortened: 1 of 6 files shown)


test/check.sh
```diff
@@ -570,6 +570,24 @@ it_can_check_with_tag_filter() {
...
+  local ref4=$(make_annotated_tag $repo "1.0-production" "tag 2")
+  local ref5=$(make_annotated_tag $repo "2.0-staging" "tag 3")
+  local ref6=$(make_commit $repo)
+  local ref7=$(make_annotated_tag $repo "2.0-staging" "tag 5")
+  local ref8=$(make_commit $repo)
+  local ref9=$(make_annotated_tag $repo "2.0-production" "tag 4")
+  local ref10=$(make_commit $repo)
+
+  check_uri_with_tag_regex $repo ".*-staging" | jq -e "
+    . == [{ref: \"2.0-staging\", commit: \"$ref6\"}]
```
xtremerui, line 587: why in here the check doesn't return `ref3` and `ref8`? what is the reasoning behind `tail -1` in the implementation? Thx.

jlsalmon: hey @xtremerui, thanks for taking the time to review! :)

I have taken the approach of duplicating the exact same logic as for the `tag_filter` option, which also behaves this way (uses [`tail -1`](https://github.com/concourse/git-resource/blob/master/assets/check#L149)) in this scenario. I've also duplicated the exact same set of tests as for `tag_filter`, which in this case also return only `ref6`. 

I can't say for certain whether this is _correct_, but I can say that it's at least _consistent_ with `tag_filter` in terms of behaviour. 

I'd be more than happy to add additional tests here if you think that would clarify the situation.

xtremerui: I think the check would only return the latest version by default so `tail -1` makes sense here. However `ref8` is actually newer than `ref6` plus it matches the regx, should `ref8` be expected other than `ref6`?

jlsalmon: correct me if I'm wrong, but `ref9`  (`2.0-production`) points to `ref8`. and `ref7` (`2.0-staging`) points to `ref6`. to me this looks correct, I would expect the regex `.*-staging` to return the latest commit that is pointed to by the `2.0-staging` tag, which is `ref6`. 

(side note: this is the same behaviour as the test directly above this one, on like 568)

xtremerui:
@vito give this PR, should we consider deperating `tag_filter` in favor of `tag_regex`? These are quit familiar to each other in terms of implemetation and testing.

vito:
@xtremerui hmm as long as it's not too expensive/risky to maintain I think it's OK to support both, mainly because the deprecation churn seems potentially more troublesome

jlsalmon:
thanks @vito and @xtremerui for taking the time to consider this patch! looking forward to hearing what you guys decide.

xtremerui (approved):
look all good to me. Thank you for the PR!

jlsalmon:
you are most welcome @xtremerui, happy to contribute - by the way, when are you planning to make the next release?

xtremerui:
Concourse releases every two weeks on a regular basis, roughly. 6.7.2 is on the way so the next release should be early December. Do you have an urgent need for this change in the next release? Note you could always create a custom resource type that points to a dev image of this resource to get the latest update.

jlsalmon:
No urgent rush, although it would be good to have this in the next release if possible. We’re already using a custom resource from our fork, and are looking to go back to the official version asap.

PR #341, opened 2020-11-16
Replace preserve rebase with merges rebase

Haegi:
According to git `--preserve-merges` is deprecated and replaced by `--rebase-merges`.
For more details on the deprecation of `--preserve-merges` see [here](https://git-scm.com/docs/git-pull#Documentation/git-pull.txt---rebasefalsetruemergespreserveinteractive).

### DIFF
--- a/assets/out
+++ b/assets/out
@@ -185,7 +185,7 @@ elif [ "$rebase" = "true" ]; then
     echo "rebasing..."
 
     git fetch push-target "refs/notes/*:refs/notes/*"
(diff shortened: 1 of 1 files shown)


xtremerui (approved):
It is quite interesting to read through the changes happening behind these two rebase options. I can't finish all but I get that `--rebase-merge` is less broken and aiming at maintain the commit topology when rebasing. 

Given all tests are passing, we should move on. Thx for the PR!

PR #346, opened 2020-12-22
Allow for returning more than just the latest version in a check

zoonage:
We have a use case in which we need to have more than just the latest commit show up from our checks to allow for a quick roll back after a deploy. by pinning a commit that is no longer in use by a resource.

This change adds a new `version_depth` parameter (I had no idea what to call it, please suggest better names). This default to the existing behaviour of taking the latest, however also provides for "getting the last X number of commits" as well.

Signed-off-by: Aaron George <aaron@ometria.com>

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -31,6 +31,7 @@ ref=$(jq -r '.version.ref // ""' < $payload)
 skip_ci_disabled=$(jq -r '.source.disable_ci_skip // false' < $payload)
 filter_whitelist=$(jq -r '.source.commit_filter.include // []' < $payload)
 filter_blacklist=$(jq -r '.source.commit_filter.exclude // []' < $payload)
+version_depth=$(jq -r '.source.version_depth // 1' < $payload)
 reverse=false
 
 configure_git_global "${git_config_payload}"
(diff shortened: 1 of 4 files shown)


AClarkie (approved):

assets/check
```diff
@@ -147,7 +148,7 @@ if [ -n "$tag_filter" ]; then
       if [ -n "$branch" ]; then
         branch_flag="--merged $branch"
       fi
-      tag=$(git tag --list "$tag_filter" --sort=creatordate $branch_flag | tail -1)
+      tag=$(git tag --list "$tag_filter" --sort=creatordate $branch_flag | tail -$version_depth)
```
vito, line 151: Does this (and the other change) need to be reversed too?

zoonage: Updated to include the bits under tag_filter. I've also added a new test for them

vito:
Looks like the `it_returns_list_of_all_tags_in_metadata` tests are failing now:

https://ci.concourse-ci.org/builds/71672458#L60064a91:4542:4594

zoonage:
> Thanks for the update! Looks like the `it_returns_list_of_all_tags_in_metadata` tests are failing now though:
> 
> https://ci.concourse-ci.org/builds/71672458#L60064a91:4542:4594

Looked like it was a random failure as `035a810` didn't really change anything but succeded

vito (approved):
weird; maybe it ran at a bad time, I've been changing the base image around. lgtm then - thanks!

PR #350, opened 2021-01-27
Added 'tag_regex_invert' option

### LINKED ISSUE #349: tag_regex: inverse match
Hi!
Currently i am facing the task of triggering a build pipeline when only for commits that don't match some tag regex.
I want to exclude in my staging environment everything that does not match `.*[0-9]+.[0-9]+.[0-9]+.*` 
Basically the `check` uses somethign like this: 
`git tag --list --sort=creatordate | grep -Ex ${regex}`
In my special case i would use invert matching using `grep -v` 
e.g. `git tag --list --sort=creatordate | grep -Exv ".*[0-9]+.[0-9]+.[0-9]+.*"`

My proposal: a bool `tag_regex_invert`that would enable the `-v` option for the grep commands using `tag_regex`.

siredmar:
This issue fixes #349.
The implementation is oriented on #339. 
If `tag_regex_invert` is set to `true` grep uses option `-v`
```
-v, --invert-match
              Invert the sense of matching, to select non-matching lines
```

Tests (check.sh) are passing.

### DIFF
--- a/test/check.sh
+++ b/test/check.sh
@@ -608,6 +608,42 @@ it_can_check_with_tag_regex() {
   "
 }
 
+it_can_check_with_tag_regex_invert_true() {
+  local repo=$(init_repo)
+  local ref1=$(make_commit $repo)
+  local ref2=$(make_annotated_tag $repo "1.0-staging" "tag 1")
+  local ref3=$(make_commit $repo)
+  local ref4=$(make_annotated_tag $repo "1.0-production" "tag 2")
+  local ref5=$(make_annotated_tag $repo "2.0-staging" "tag 3")
+  local ref6=$(make_commit $repo)
+  local ref7=$(make_annotated_tag $repo "2.0-staging" "tag 5")
+  local ref8=$(make_commit $repo)
+  local ref9=$(make_annotated_tag $repo "2.0-production" "tag 4")
+  local ref10=$(make_commit $repo)
+
+  check_uri_with_tag_regex_invert $repo ".*-staging" true | jq -e "
+    . == [{ref: \"2.0-production\", commit: \"$ref8\"}]
+  "
+}
+
+it_can_check_with_tag_regex_invert_false() {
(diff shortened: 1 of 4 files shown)


test/check.sh
```diff
@@ -652,6 +688,50 @@ it_can_check_with_tag_regex_with_cursor() {
...
+  local ref8=$(make_annotated_tag $repo "2.0-production" "tag 4")
+  local ref9=$(make_commit $repo)
+  local ref10=$(make_annotated_tag $repo "3.0-staging" "tag 5")
+  local ref11=$(make_commit $repo)
+  local ref12=$(make_annotated_tag $repo "3.0-production" "tag 6")
+  local ref13=$(make_commit $repo)
+
+  x=$(check_uri_with_tag_regex_invert_from $repo ".*-staging" true "2.0-staging")
+  check_uri_with_tag_regex_invert_from $repo ".*-staging" true "2.0-staging" | jq -e "
+    . == []
```
xtremerui, line 709: I think this one here should return 2.0-production and 3.0-production

dji77: https://bitbucket.org/product/guides/basics/four-starting-steps?utm_source=engagement-engine&utm_medium=email&utm_campaign=4-starting-steps_EML-6176&ep_click_type=primary

siredmar: @dij77: what are you trying to tell?

siredmar: @xtremerui are you sure? If this is true, maybe i need an explanation about the `cursor`.

siredmar: @xtremerui you are right. I try to fix this.

xtremerui:
@siredmar just following up, are you still working on this PR?

siredmar:
@xtremerui indeed not. Moved along as this feature is not needed anymore. Feel free to continue this. Otherwise feel free to close the PR.

xtremerui:
I will leave the issue open and close this PR.

PR #359, opened 2021-05-12
Fix #358 Add support for passphrase protected private key files.

### LINKED ISSUE #358: Support password protected private key files
https://github.com/concourse/git-resource/pull/10 added a clear error message when the private key is password protected.
The issue is that the private key is written to disk inside the container, and any developers with access to fly intercept would then be able to read the private key for the service account. 

For an open source org it's not the end of the world. You can have a private key that has read-only access to the repo as long as the workflow doesn't require write access.

Is there a good reason to avoid supporting this workflow? I think it's a necessity for my use case.

jamie-pate:


vito (approved):
looks good - thanks!

PR #360, opened 2021-05-16
support proxy for http(s) URIs

abetaev:
i found that specifying `https_tunnel` does not work with HTTP(S) URIs, btw it makes git-resource fail on missing file (`.ssh` directory does not exist for HTTP(S) URIs).

apparently `HTTPS tunnel` is SSH specific terminology, that's why i used separate section of `source` to provide proxy information. if needed username and password can be provided within URL itself. almost all Linux apps respect these environment variables as system-wide proxy configuration.

i found no way of providing environment to resources using Concourse. for me it seems that providing environment from Concourse would be more logical (and flexible) rather than making each plugin to support specific features to set environment. Concourse allows to do it for tasks; are resources that different?

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,12 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
   * `proxy_password`: *Optional.* If the proxy requires authenticate,
       use this password
 
+* `proxy`: *Optional.* Information about proxy to be used for communicating
+  Has the following sub-properties:
+  * `http_proxy`: *Optional.* Proxy for HTTP communication, for example `http://proxy.example:8080/`
+  * `https_proxy`: *Optional.* Proxy for HTTPS communication for example `socks://user:secr3t@proxy.example:3128/`
+  * `no_proxy`: *Optional.* Hosts to be communicated directly, for example `localhost,.mydomain.com`
+
 * `commit_filter`: *Optional.* Object containing commit message filters
   * `commit_filter.exclude`: *Optional.* Array containing strings that should
     cause a commit to be skipped
--- a/assets/check
+++ b/assets/check
@@ -17,6 +17,7 @@ cat > $payload <&0
 
 load_pubkey $payload
 configure_https_tunnel $payload
+configure_proxy $payload
 configure_git_ssl_verification $payload
 configure_credentials $payload
 
--- a/assets/common.sh
+++ b/assets/common.sh
@@ -60,6 +60,19 @@ EOF
   fi
 }
 
+configure_proxy() {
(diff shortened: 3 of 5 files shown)


taylorsilva:
Hey, thanks for the PR!

Concourse workers will actually pass in the `http_proxy` and `https_proxy` env var's into containers if they're set on the host. https://github.com/concourse/concourse/blob/71743c481c769c5b238d9dee80e9766f9c33faec/atc/worker.go#L15-L16

With this in mind, is this PR still needed to get this working for you?

abetaev:
@taylorsilva it didn't work for me, i tried to set this on worker host first; all Concourse workers (and web) have `http_proxy`, `https_proxy` and `no_proxy` environment variables set;

i didn't figure how to check environment of resource container, but it started to work only when i explicitly made this change.

taylorsilva:
Huh that's strange. I quickly verified with a local docker-composed concourse that setting all three env var's on the worker host and seeing that they were passed into check, get, and task containers.

```
└─▪ fly -t l i -u http://localhost:8080/teams/main/pipelines/test/jobs/simple-job/builds/1
1: build #1, step: image, type: check, expires in:
2: build #1, step: image, type: get
3: build #1, step: simple-task, type: task
choose a container: 1
bash-5.1# env | grep proxy
no_proxy=example.com
https_proxy=someproxy.com
http_proxy=someproxy.com

└─▪ fly -t l i -u http://localhost:8080/teams/main/pipelines/test/jobs/simple-job/builds/1
1: build #1, step: image, type: check, expires in:
2: build #1, step: image, type: get
3: build #1, step: simple-task, type: task
choose a container: 2
bash-5.1# env | grep proxy
no_proxy=example.com
https_proxy=someproxy.com
http_proxy=someproxy.com

└─▪ fly -t l i -u http://localhost:8080/teams/main/pipelines/test/jobs/simple-job/builds/1
1: build #1, step: image, type: check, expires in:
2: build #1, step: image, type: get
3: build #1, step: simple-task, type: task
choose a container: 3
bash-5.1# env | grep proxy
no_proxy=example.com
https_proxy=someproxy.com
http_proxy=someproxy.com
```

Are you not seeing the same? If you already have a setup for testing the resource you could try adding a debug line to print `env` and see if any of the proxy env vars are being passed in.

abetaev:
yeah, strange, i just checked check container of type docker-image resource and it had those env variables. i will check with git-resource again bit later. could i was doing something wrong :-/

abetaev:
the issue was unrelated to this pull-request
in my case there is something wrong with image fetching from public hub.docker.com

PR #361, opened 2021-06-07
Avoid writing resource stdinput to disk. It makes it too easy to 'accidentally' see credentials etc.

jamie-pate:


vito (approved):
awesome, thanks!

PR #363, opened 2021-07-08
Allow configuring `source.ref` to track a named ref

aoldershaw:
`source.ref` is used to track a named ref in a git commit. it is intended as an alternative to `source.branch` when the
thing you want to track isn't a branch.

The primary motivation here is to track GitHub pull requests using the ref name `pull/ID/head`

check and get both fetch the single ref by:
1. initializing an empty git repo
2. adding the uri as a remote
3. fetching the ref

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,21 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
 
 * `uri`: *Required.* The location of the repository.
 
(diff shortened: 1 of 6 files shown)


aoldershaw:
By the way @chenbh @EstebanFS I'm thinking about taking a different approach for tracking PRs (not using the git-resource). Not sure what other use cases there are for this feature, so I could probably just close this

PR #366, opened 2021-07-22
Added committer_author to the outputs for use in slack notifications

jdziat:
As part of our workflow we use the .git directory files to populate out slack notifications. Instead of running a secondary command or task we'd like the git resource to output the author name into a separate file. This way we can avoid instances of unusable or appended emails from being sent to our notification channels.


jdziat:
@taylorsilva Any chance we can get this merged into the branch? It's a simple one liner that would help us a lot.

taylorsilva (approved):
LGTM

We assign PR's to someone every morning during our standup, so you shouldn't have to worry about us missing a PR :)

PR #371, opened 2021-09-02
Add the ability to push to refs other than refs/heads

jamie-pate:
Some workflows do not follow the default git reference conventions, rarely you may want to e.g. create a review in gerrit, which requires a push to `refs/for/branch` where you would normally push to `refs/heads/branch`.

See https://gerrit-review.googlesource.com/Documentation/user-upload.html#_git_push


jamie-pate:
unfortuately the get step then fails with 

`fatal: reference is not a tree: 6c8f542428b6d81e7a52027539cedc4639476de7`

where `6c8f542428b6d81e7a52027539cedc4639476de7` is the commit that was pushed...

taylorsilva (approved):
LGTM - not sure what other worklfows this supports, but maybe someone will find it useful :)

PR #372, opened 2021-09-03
Add support for pulling from a previously pushed gerrit review commit

jamie-pate:
This compliments the previous pull request https://github.com/concourse/git-resource/pull/371

In a Gerrit code review workflow the user pushes changes to `refs/for/<branchname>` and then the Gerrit server will create a ref at `changes/nnnnn[/m]` for code review. 

Unfortunately, the default put/get combination that concourse uses does not work as it expects the commit that was pushed to be present on the branch immediately.

This commit will opt in to the ability to search the remote refs for the same version that was pushed during the `put` step and check it out if it is a ref advertised by the remote, satisfying the `put` + `get` mechanism of concourse. 

See https://gerrit-review.googlesource.com/Documentation/user-upload.html#_git_push a description of the Gerrit behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,12 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
 
 * `version_depth`: *Optional.* The number of versions to return when performing a check
 
+* `search_remote_refs`: *Optional.* True to search remote refs for the input version when checking out during the get step.
+    This can be useful during the `get` step after a `put` step for unconventional workflows. One example workflow is the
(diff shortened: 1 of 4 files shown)


jamie-pate:
Oops, It looks like this falls in the same category as pushing to a branch:

```
Note that the version produced by the put step will be picked up by subsequent get steps even if the branch differs from the branch specified in the source. To avoid this, you should use two resources of read-only and write-only.
```

taylorsilva (approved):

PR #380, opened 2022-03-29
workaround checkout-error for nested shallow submodules

ccwienk:
If a nested submodule is configured to be "shallow", referencing a
commit that is not reachable from the respective repository's HEAD
(typically the default branch, if e.g. hosted on GitHub), will result in
an error when checking out the submodule (git submodule update..).

Workaround this by always setting `--no-recommend-shallow`, which
lets git ignore this setting.

For a "real-world" example, see: https://github.com/google/docsy/issues/755 (I also commented there)

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -176,7 +176,7 @@ if [ "$submodules" != "none" ]; then
       continue
     fi
 
-    git submodule update --init --no-fetch $depthflag $submodule_parameters "$submodule_path"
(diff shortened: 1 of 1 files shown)


xtremerui:
My knowledge with git shallow clone is limited. Would adding this flag cause users cloning more data than they want (if their submodules is meant to be shallow)?

taylorsilva:
Closing because stale. Anyone is free to re-open this or make a new PR if they want to.

PR #385, opened 2022-06-03
Mention the regex sensitivity of `commit_filter` strings

jgriff:
Polished the README to note that the `commit_filter` option is
sensitive to regex characters.  This was not immediately obvious and can
cause a bit of confusion.  Even leading to suspecting a defect in the
implementation of the filter as it can send the resource into a whacky
state of not selecting any commits (since the filter expression would
then be malformed).

Updated the README to mention this, and give an example of configuring
the `commit_filter.exclude` with a bracketed string that escapes the
bracket characters.

```yaml
  commit_filter:
    exclude: ["\\[skip deploy\\]", "\\[deploy skip\\]"]
```

Lastly, also fixed a typo and removed the fully qualified dot notation
when describing the sub-elements of `commit_filter` in the README.

was:

```yaml
  commit_filter:
    commit_filter.exclude:
    commit_filter_include:
```

now:

```yaml
  commit_filter:
    exclude:
    include:
```

This aligns with the convention used in all other configuration options.

Only the README was changed in this commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,11 +113,19 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
       use this password
 
 * `commit_filter`: *Optional.* Object containing commit message filters
-  * `commit_filter.exclude`: *Optional.* Array containing strings that should
+  * `exclude`: *Optional.* Array containing strings that should
     cause a commit to be skipped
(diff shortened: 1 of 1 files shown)


xtremerui (approved):
LGTM. Thank you for the PR!

Issue #389, opened 2022-08-31
CommitFilter --all-match


jakubikan:
We are currently trying to filter out commits via the commit filter.

However looking into the code when add multiple keywords it will filter all commits containing the keywords with an "OR". 

the `git rev-list` command provides an option `--all-match` to match all greps to filter the commits with an "AND"
https://git-scm.com/docs/git-rev-list#Documentation/git-rev-list.txt---all-match

Would it possible to add this via property?

PR #390, opened 2022-09-20
Introducing include_all_match/exclude_all_match

jakubikan:
I have submitted an Issue where we had the Problem that the commit-filter matches via --grep but adding multiple to items to the commit filter will add more --greps to the filter whitelist 
However we need match these multiple greps via an AND not OR

This would be the PR to use the --all-match described here https://github.com/concourse/git-resource/issues/389


jakubikan:
Hmm thats strange ubuntu did not pass but alpine did

taylorsilva:
Replaced by #432 
Thanks for the PR!

Issue #325, opened 2020-06-22
branch parameter allows passing additional flag to git clone


micw-dec:
While experimenting with different branches from a git repository we noticed that it's possible to pass an additional flag to `git clone` via the `branch` parameter in the source configuration, for example:

```
resources:
  - name: some-repo
    type: git
    source:
      uri: some-uri
      branch: some-branch --no-single-branch
```

In that case the `--single-branch` option passed to `git clone` in the in-script is overridden, it looks like the branch-parameter is not checked before passing it on.

Maybe one possible solution could be to split the parameter on space and ignore the rest (or exit with an error code?).

jforand:
I would actually like to vote against this being impemented. 
This undocumented feature is a life saver for our team to be able to perform a `git diff` between different branches in the same repo

taylorsilva:
Not changing since it seems like folks rely on this. yay for unintended features! 🫠

PR #393, opened 2022-12-02
Add 'all_branches' flag to params in 'in'

multimac:
Adds a flag to disable the use of the '--single-branch' flag when cloning the repository, allowing all branches to be fetched from the remote

Removes the need for workarounds such as passing `--no-single-branch` via `source.branch`, as described in #325


taylorsilva:
@multimac any interest in picking this back up? Would be happy to get it merged in.

Issue #265, opened 2019-07-11
Ability to specify maximum number of retries?


EduardoAC:
Recently, I have this problem during rebase that it's causing an infinite loop 

Is there a way to specify a "max retry" count? Or some other way to have the job fail after a certain amount of time?

![Screenshot 2019-07-11 at 10 15 45](https://user-images.githubusercontent.com/2104706/61043414-1c9aac80-a3ce-11e9-9549-a9e6b792498c.png)

Thank you for your time, I appreciate it.

ukabu:
We have the same issue. It really should fail after a few retries (there should be a default max number of retries).

spion:
After submitting a PR for this, I realized concourse already has a [`timeout` option](https://concourse-ci.org/timeout-step.html) which will stop the step after some time. In the absence of retries I think that should do the trick.

taylorsilva:
Workaround provided by @spion to use the `timeout` step modifier. Closing this issue since it's not worth fixing since `timeout` will get you the same behaviour, albeit slightly less robust.

EduardoAC:
@taylorsilva It makes sense; thanks for letting me know

PR #395, opened 2023-01-18
feat: limit retries

spion:
Hi there! This PR adds a limit on the retries of rebase and merge.

Fixes https://github.com/concourse/git-resource/issues/265

Due to how dangerous infinite retries can potentially be if something goes wrong, I took the liberty of also changing the default to 10 retries. Do you think that's ok?

I would appreicate if you have any guidance about implementing tests. I'm not sure how to hook into the loop and make an arbitrary push fail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ Tracks the commits in a [git](http://git-scm.com/) repository.
     *MUST* be included in commit messages for the commit to not be
     skipped
 
-  **Note**: *You must escape any regex sensitive characters, since the string is used as a regex filter.*  
+  **Note**: *You must escape any regex sensitive characters, since the string is used as a regex filter.*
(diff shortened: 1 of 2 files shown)


taylorsilva (changes requested):
Overall implementation LGTM
It feels like this could have a test added to it though to ensure we don't regress on #265 in the future.

Not sure if you're still around to add a test. I have some time right now so I'll try and add one. I'll push something to the branch if I'm happy with it.

taylorsilva:
Just read your comment here: https://github.com/concourse/git-resource/issues/265#issuecomment-1386969933
and yeahhh that's a good point, that you can just use `timeout` on the put step. Gonna say adding this is not worth the effort of maintenance. Going to close the parent issue as well since the workaround is valid imo. Thanks for the PR though! Appreciate the effort you put into figuring it out.

PR #397, opened 2023-01-30
Add options out parameter.

romain-dartigues:
This parameter let the Concourse resource to set the `--push-option`.

* https://git-scm.com/docs/git-push#Documentation/git-push.txt---push-optionltoptiongt
* https://gitlab.com/gitlab-org/gitlab/-/blob/v15.8.0-ee/doc/user/project/push_options.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,6 +322,9 @@ and the `rebase` parameter is not provided, the push will fail.
 
 * `only_tag`: *Optional.* When set to 'true' push only the tags of a repo.
 
+* `options`: *Optional.* When not null, add it to the
+  [`--push-options`](https://git-scm.com/docs/git-push#Documentation/git-push.txt---push-optionltoptiongt).
+
 * `tag_prefix`: *Optional.* If specified, the tag read from the file will be
 prepended with this string. This is useful for adding `v` in front of
 version numbers.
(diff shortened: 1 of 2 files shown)


romain-dartigues:
(review requested, especially do you think it should support multiple options?)

taylorsilva:
What's the usecase you imagine for this? I mechanically understand what you're adding here, but I don't know _when_ I would use this? Do you have an example?

romain-dartigues:
Hi Taylor,
my apologies, I haven't seen the notification.

One obvious use-case would be the one in the GitLab link, ie: `--push-option=ci.skip` (and/or add add variables required to trigger some parts of the pipeline through `-o ci.variable="foo=bar"`).

In my case, one of the project hosted on a GitLab use a mix of Concourse pipelines and GitLab CI pipelines; being able to disable or trigger the GitLab pipeline trough the `git-resource` of Concourse would be useful.

taylorsilva (changes requested):
Sorry for the delay, lots of stuff to look at across all the repos!

Noticing we have no test here for this and we should add one I think: https://github.com/concourse/git-resource/blob/master/test/put.sh

taylorsilva:
Closing because stale. Feel free to re-open/re-submit with added tests.

PR #401, opened 2023-05-03
Feature/reject invalid filter format

fenech:
I ran into this issue when I accidentally passed a single string to the "commit_filter.include" parameter. I was confused by the fact that the check appeared to work, but it returned the wrong commit.

Instead, this makes the check fail when the configuration of this parameter is incorrect.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -191,7 +191,7 @@ ADD test/ /tests
 RUN /tests/all.sh
 
 FROM resource AS integrationtests
-RUN apt update && apt install -y squid
+RUN apt update && apt install -y net-tools squid
 ADD test/ /tests/test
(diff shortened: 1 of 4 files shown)


taylorsilva:
Awesome enhancement! I'll review and merge or give feedback. At a quick glance, everything looks pretty good!

fenech:
Thanks for the initial feedback @taylorsilva! Did you get a chance to look at it in the end?

taylorsilva (commented):
Sorry for the delay @fenech IRL stuff pulled me away and I'm just getting back in the habit of checking stuff here. This PR is solid! Nice little improvements to the areas you had to touch and added a test too. THANK YOU for the PR!

PR #405, opened 2023-05-18
fix gitlab 16.0 legacy urls removed

m-roskrow:
as of gitlab 16.0 the commit URL no longer supports 'legacy' links: https://docs.gitlab.com/ee/update/deprecations.html#legacy-urls-replaced-or-removed

the dash is now required to properly link to a commit, otherwise a 404 is returned


taylorsilva (approved):
LGTM

PR #418, opened 2023-11-21
Optimize check

rmg:
YMMV, but with a large monorepo we saw:
 - 99% reduction in network traffic between the concourse workers and GHE server
 - 90% reduction in CPU usage on the server during clone/fetch

#### check: use bare clone

The check container doesn't make any use of an actual checkout, but on
large repositories the time to checkout after fetching can be 90% or
more of the overall time to clone.

Because the `--bare` option prevents the `--branch` option from being
setting the refspec in the config we have to do it as a follow-up step.

Instead of relying solely on the config having the refspec set
correctly, we can also explicitly specify the branch when fetching if
one is set.

#### check: use blobless partial clones

The check operation doesn't need any blobs. The most detail is ever
needs about commits is the paths they touched, which only requires
commit and tree objects, not the blobs.

On large repositories, especially with lots of history, this can
dramatically reduce the time to clone and fetch while also dramatically
reducing the server-side resources used to serve up those large git
repositories.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -56,7 +56,7 @@ else
     branchflag="--branch $branch"
   fi
 
-  git clone --bare --single-branch $uri $branchflag $destination $tagflag
+  git clone --bare --filter=blob:none --single-branch $uri $branchflag $destination $tagflag
   cd $destination
   # bare clones don't configure the refspec
   if [ -n "$branch" ]; then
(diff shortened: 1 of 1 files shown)


taylorsilva (approved):
LGTM - Thanks for the PR and this AMAZING improvement!

I just merged another PR so I rebased your PR locally on `master` and ran the tests again to be sure everything still worked as expected. Everything so passed ✅

marco-m-pix4d:
@rmg we went through a similar pain with big repos. HUGE thanks!

Issue #267, opened 2019-07-29
Suggestion: Allow git sparse checkout / checkout specific directory


wyattjsmith1:
As I have everything in a monorepo, it would be great if we could checkout specific directories as resources. 

Consider a case with a monorepo with 2 services and a `ci` directory for task configurations:
```
/
|- ci
|- service-a
|- service-b
```
 I only want to run tests on a specific service if the files within that service change, so I'd use two separate resources (named `service-a/b`) with the `paths` param set to their respective directories:
```
- type: git
  name: service-a
  source:
    uri: "https://my-repo"
    paths:
    - service-a/**
- {service b definition here}
- type: git
  name: ci
  source:
    uri: "https://my-repo"
    paths:
    - ci/**
```
 I'd have two jobs, one for each service, and each job would start with a `get: service-a/b` and a `get: ci`:
```
- name: build-service-a
  plan:
  - get: service-a
    trigger: true
  - get: ci
  - task: build
    file: ci/ci/task.yaml
```
This is great, but it doesn't actually clone `service-x` and `ci` into the task working directory, it clones the entire repository. In the case above, the task directory would contain two instances of the monorepo, just named differently. There is no technical issue with this, but it is annoying and confusing to have to prepend `/ci/ci` when specifying my task.

I believe allowing a git resource to clone a sub-directory of the repository would solve this issue easily.

stwomack:
Bump. @wyattjsmith1 Did you ever get an response on this?

wyattjsmith1:
Nope! I don't use concourse anymore, but I still think this would be a nice feature.

PR #427, opened 2024-04-12
Add ability to perform a sparse checkout

caseybraithwaite:
Hi 👋 ,

Haven't raised a PR on this repository before so apologies if I've done something wrong.

This PR adds the ability to do a sparse checkout.

There was some consideration in using the existing `.source.paths` as the paths to checkout but, within our pipelines, we came across a scenario where we wanted to trigger on certain paths but then checkout additional paths. Hopefully this is acceptable for other people.

This should close [267](https://github.com/concourse/git-resource/issues/267).

### DIFF
--- a/assets/in
+++ b/assets/in
@@ -33,6 +33,7 @@ configure_credentials "$payload"
 
 uri=$(jq -r '.source.uri // ""' <<< "$payload")
 branch=$(jq -r '.source.branch // ""' <<< "$payload")
+sparse_paths="$(jq -r '(.source.sparse_paths // ["."])[]' <<< "$payload")" # those "'s are important
 git_config_payload=$(jq -r '.source.git_config // []' <<< "$payload")
 ref=$(jq -r '.version.ref // "HEAD"' <<< "$payload")
 override_branch=$(jq -r '.version.branch // ""' <<< "$payload")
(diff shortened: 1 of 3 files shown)


xtremerui:
Hi thx for the PR. Could you add a test here https://github.com/concourse/git-resource/blob/05da2027655664970f648ebc01f010e42f4f5ee9/test/get.sh ?

taylorsilva:
Thanks for adding tests @caseybraithwaite 
This is in my review backlog so I'll get to it eventually.

taylorsilva (approved):
Sorry for the VERY long wait. I'm never too keen to peer into these bash scripts. This is a great feature though!

Only thing missing is documenting the new field in the README. I'll do that after merging though.

I compiled and locally tested out this resource; it works as expected!

```yaml
resource_types:
  - name: git
    type: registry-image
    source:
      repository: taylorsilva/git-resource
      tag: latest

resources:
  - name: repo
    type: git
    source:
      uri: https://github.com/concourse/concourse.git
      paths:
        - atc/**
      sparse_paths:
        - atc/**
        - worker/**

jobs:
  - name: job
    plan:
      - get: repo
      - task: ls
        config:
          platform: linux
          image_resource:
            type: mock
            source:
              mirror_self: true
          inputs:
            - name: repo
          run:
            path: sh
            args:
              - -c
              - |
                ls -lh ./repo
```

PR #324, opened 2020-06-09
bug: fixed an edge case bug.

evanchaoli:
Fixes: #323

Signed-off-by: Chao Li <chaol@vmware.com>

Issue #323, opened 2020-06-09
[bug] Doesn't pick up latest commit


evanchaoli:
One of our users reported an issue today where he complained his latest commit was not picked up by the `git` resource.

I looked into the issue, and found:

1. The `git` resource used `paths` to specify a file to monitor.
2. It held a single version, where the version is the first commit of the git repo.
3. The user pushed a new commit that modified the file, so the new commit was supposed to be picked up, but not.

I managed to find that `git` resource uses the following command to list target commits:

```
$ git rev-list --all --first-parent 7a4617bae490f29513d97ea7f7538fe64c03bf94~1..HEAD -- <path-to-monitor>
fatal: bad revision '7a4617bae490f29513d97ea7f7538fe64c03bf94~1..HEAD'
```
Where `7a4617bae490f29513d97ea7f7538fe64c03bf94` is the first commit of the repo, so `7a4617bae490f29513d97ea7f7538fe64c03bf94~1` is invalid.

Then I thought `git` resource should check if or not a commit is the first commit, then I found the check is there already: https://github.com/concourse/git-resource/blob/master/assets/check#L64:

```
 64   init_commit=$(git rev-list --max-parents=0 HEAD)
 65   if [ "${ref}" = "${init_commit}" ]; then
```

but this statement is not perfect. In the user's repo, it returns two commits:

```
root@fc791363-5549-48e9-6814-d61c8163f5dc:/tmp/git-resource-repo-cache# git rev-list --max-parents=0 HEAD
5f003be10a4b905143390e5fb7a9389660648554
7a4617bae490f29513d97ea7f7538fe64c03bf94
```
So at line 65, `ref` didn't equal to `init_commit`.

So far, the fix is clear and simple, we should only take the last line of `git rev-list` at line 64 returns multiple commits:

```
 64   init_commit=$(git rev-list --max-parents=0 HEAD | tail -1)
```

P.S. I also tried `git rev-list --max-parents=0 HEAD -n 1 --reverse`, which didn't work. The command seemed to pick n comments first then reverse the order, so it still didn't return the first commit.

PR #433, opened 2025-04-04
handle non-initial orphan commits as current resource version

mstandley-tempus:
If the git resource has a parentless commit as its most recent resource version, the check command fails with the error:

```
$ git rev-list --all --first-parent abc123~1..HEAD
fatal: ambiguous argument 'abc123~1..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

Most typically, a parentless commit will be the initial commit in a branch, and this is already handled today.  But if the parentless commit is not the initial commit, this is not handled, and concourse will stop seeing newer commits on that branch.  These non-initial parentless commits are rare, but possible.  See: https://stackoverflow.com/a/69829363

The workaround for this situation is to make sure the most recent commit on the branch actually has a parent, then delete the resource versions with `fly clear-versions`.

So, rather than getting stuck and requiring manual intervention, let's teach the git-resource to recover from this situation on its own by avoiding `~1` when we know `${ref}` is without a parent.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -103,14 +103,18 @@ else
 fi
 
 if [ -n "$ref" ] && git cat-file -e "$ref"; then
-  init_commit=$(git rev-list --max-parents=0 HEAD | tail -n 1)
-  if [ "${ref}" = "${init_commit}" ]; then
-    reverse=true
-    log_range="HEAD"
-  else
-    reverse=true
-    log_range="${ref}~1..HEAD"
-  fi
+  reverse=true
+  log_range="${ref}~1..HEAD"
+
+  # if ${ref} does not have parents, ${ref}~1 raises the error: "unknown revision or path not in the working tree"
+  # the initial commit in a branch will never have parents, but rarely, subsequent commits can also be parentless
+  orphan_commits=$(git rev-list --max-parents=0 HEAD)
+  for orphan_commit in ${orphan_commits}; do
+    if [ "${ref}" = "${orphan_commit}" ]; then
+      log_range="HEAD"
+      break
+    fi
+  done
 else
   log_range=""
   ref=""
--- a/test/check.sh
+++ b/test/check.sh
@@ -956,6 +956,28 @@ it_can_check_a_repo_having_multiple_root_commits() {
   "
 }
 
+it_can_check_a_repo_having_multiple_root_commits_from_the_orphan_commit() {
+  local repo=$(init_repo)
+  local ref1=$(get_initial_ref $repo)
+  local ref2=$(make_commit $repo)
+
+  # Same as above, but where the current version is the orphan commit.
+  git -C $repo checkout --orphan temp $ref2
+  git -C $repo commit -m "second root" --allow-empty
+  git -C $repo checkout master
+  git -C $repo merge temp --allow-unrelated-histories -m "merge commit"
+  second_root=$(git -C $repo rev-parse HEAD^2)
+  ref3=$(git -C $repo rev-parse HEAD)
+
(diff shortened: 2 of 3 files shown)


mstandley-tempus:
Compare: https://github.com/concourse/git-resource/issues/323 and https://github.com/concourse/git-resource/pull/324

Same issue where there are multiple commits without parents on a branch.  This PR extends that fix by also handling the situation where that non-initial parentless commit appears as the latest version.  It adds a test covering this scenario

```
running it_can_check_a_repo_having_multiple_root_commits_from_the_orphan_commit...
  Switched to a new branch 'temp'
  [temp (root-commit) b8540e3] second root
   1 file changed, 1 insertion(+)
   create mode 100644 some-file
  Switched to branch 'master'
  Merge made by the 'ort' strategy.
  false
  false
  Cloning into bare repository '/tmp/git-tests.XPkmqf/git-tests.j7vQP1/git-resource-repo-cache'...
  warning: --filter is ignored in local clones; use file:// instead.
  done.
  false
  false
  fatal: ambiguous argument 'b8540e3299bb941bd6c8f7cbb3ca6d39c43b0743~1..HEAD': unknown revision or path not in the working tree.
  Use '--' to separate paths from revisions, like this:
  'git <command> [<revision>...] -- [<file>...]'
  []
  false
```

With the fix:

```
running it_can_check_a_repo_having_multiple_root_commits_from_the_orphan_commit...
  Switched to a new branch 'temp'
  [temp (root-commit) f464d3c] second root
   1 file changed, 1 insertion(+)
   create mode 100644 some-file
  Switched to branch 'master'
  Merge made by the 'ort' strategy.
  false
  false
  Cloning into bare repository '/tmp/git-tests.dG22US/git-tests.arTZR3/git-resource-repo-cache'...
  warning: --filter is ignored in local clones; use file:// instead.
  done.
  false
  false
  [
    {
      "ref": "2c2f129b5f9b185f3fec253875fcc1b4731005d2"
    },
    {
      "ref": "477f6cb50b590182b2555c169dc46bcfa3a484d6"
    },
    {
      "ref": "7acc47fe4a4f9f0cbb0cef4a42878687996203dc"
    }
  ]
  true
```

taylorsilva:
Thanks for the PR! I'll circle back to this one in a few days probably.

taylorsilva (approved):
Thanks for adding a test as well. Made it very easy to review this PR!

mstandley-tempus:
There was still something bugging me about this pull-request, because the presence of `--first-parent` in the `git rev-list` call should mean that the orphan commit can never appear in the resource version list.  From the manual:

```
       --first-parent
          When finding commits to include, follow only the first parent commit upon seeing a merge commit. This option can give a better overview when viewing the evolution of a particular topic branch, because merges into a topic branch tend to be only about adjusting to updated upstream from time to time, and this option allows you to ignore the individual commits brought in to your history by such a merge.
```

And if I hadn't seen it happen in a real concourse instance, I would be dubious about this even being possible.  I suspect it occurs in the following scenario:

1. user wants to merge in a unrelated branch (maybe from a different repo) and force pushes, but user accidentally clobbers the branch and replaces it with a single init commit
2. check runs, new init commit appears as the latest in the concourse resource version list
3. user restores the clobbered branch locally, then merges with the unrelated branch with `--allow-unrelated-histories`, causing the same orphan commit with the same SHA to appear on that branch, then force pushes again
4. check runs, but fails because it's hitting the error case described above

The edgiest of the edge cases!  I really had to use my imagination to explain how this might have happened.

PR #438, opened 2025-04-10
Switch to wolfi-base as the base image

taylorsilva:
Related to https://github.com/concourse/concourse/issues/9183

We want to start compiling the resource types for AMD64 and ARM64 arch's. The Paketo images only release AMD64 images. Chainguard seems like a good alternative right now and should satisfy any security concerns users have.

The newer version of git changed some error messages which broke a few tests.

I moved proxytunnel into its own build stage so we don't carry over dependencies that were only required during its build.

Locally on my mac, the wolfi-base version of the resource comes in at 123MB. Cross-compiling the Paketo-Jammy version results in a 598MB image locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,187 +1,55 @@
-ARG base_image
+ARG base_image=cgr.dev/chainguard/wolfi-base:latest
 
-FROM ${base_image} AS resource
-USER root
-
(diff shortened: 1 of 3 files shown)


taylorsilva:
Keeping this as a draft for now because I'll have to change the resource's build pipeline before the checks will work.

taylorsilva:
Ran tests locally and they all passed.

PR #442, opened 2025-08-13
Fix population of .git/branch file

paul-szymanski-a9s:
When the checked out branch has a tag, the file was populated with the tag instead of the branch name.


taylorsilva (approved):
Great catch! Thanks for the PR

Issue #268, opened 2019-08-20
Can we add --force-with-lease feature?


WxFang:
We need force push to automatically rebase master branch to feature branch. 
But there can be a race condition that before force push, someone pushed some commit to the feature branch. Then force push will ignore that specific commit. 
It seems like --force-with-lease can solve this problem.

taylorsilva:
Closing, but PR's are welcome if someone wants to add this.

PR #447, opened 2025-09-04
feat: add support for `--force-with-lease`

Jonnobrow:
Adds support for `--force-with-lease`, which can be use in scenarios where `--force` is unsafe. Implements #268


Jonnobrow:
> @Jonnobrow Thanks for the PR. Is there a reason you decided to close it?

While the code successfully passes the `--force-with-lease` flag to the correct commands - I was having difficulty getting the tests to pass. Still looking into why this is, and will re-open the PR when I can get them to pass :)

PR #449, opened 2025-09-18
Add --progress flag to git clone for better CI visibility

analytically:
Include progress output during git clone operations to provide real-time feedback in CI pipelines. This helps debug slow or stalled clones and improves observability of the check command.


taylorsilva (approved):
LGTM

image

PR #452, opened 2025-09-19
Optimize partial clone filter strategy based on path filtering requirements

analytically:
When checking for new versions in large repositories, the git-resource currently uses --filter=blob:none unconditionally. However, when path filtering is not required, we can use the more aggressive --filter=tree:0 for significant performance improvements.

Changes:
- Use --filter=tree:0 when no path filtering is configured (paths="." and no ignore_paths)
- Continue using --filter=blob:none when path filtering is active, as git rev-list needs tree objects to evaluate path specifications

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -125,7 +125,16 @@ else
     branchflag="--branch $branch"
   fi
 
-  git clone --bare --filter=blob:none --single-branch --progress $uri $branchflag $destination $tagflag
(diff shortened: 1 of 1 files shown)


analytically:
This optimization automatically helps subsequent fetches too! Once a repo is cloned with --filter=tree:0, all future fetches in that cached repo will also skip trees.

taylorsilva (approved):
LGTM!

PR #461, opened 2025-12-02
fix(check): find tags on unmerged branches when tag filtering enabled

### LINKED ISSUE #448: Fetching new tags doesn’t work when tag filtering is enabled (sometimes)
### Summary

When using tag filtering with a git resource, new tags are not being fetched. The resource seems to ignore newly pushed tags unless filtering is disabled and enabled again. In the example below the tag filtering is enabled:

But it doesn't see this tag: https://github.com/aptos-labs/aptos-core/releases/tag/aptos-node-v1.34.4-hotfix even then triggered manually. The issue is not isolated to this specific repo but happens from time to time to any git resouce that has tag filtering enbabled.

Deployment type: BM server with Ubuntu 22.04.5 LTS

Image

### Steps to Reproduce

```
---
resources:
  - name: aptos-repo
    type: git
    source:
      uri: https://github.com/aptos-labs/aptos-core.git
      fetch_tags: true
      version_depth: 500
      tag_filter: "aptos-node*"
```

### Expected Results

_No response_

### Actual Results

_No response_

### Web Node(s) configuration

_No response_

### Worker(s) configuration

worker.env:
```
CONCOURSE_WORK_DIR=XXX
CONCOURSE_TSA_HOST=127.0.0.1:2222
CONCOURSE_TSA_PUBLIC_KEY=XXX
CONCOURSE_TSA_WORKER_PRIVATE_KEY=XXX
CONCOURSE_RUNTIME=containerd
CONCOURSE_CONTAINERD_ALLOW_HOST_ACCESS=true
CONCOURSE_BAGGAGECLAIM_DRIVER=overlay
CONCOURSE_CONTAINERD_DNS_SERVER=XXX
```

### Concourse Version

7.14.0

### Browser (if applicable)

_No response_

### Did this use to work?

No, 7.11.2 also had the same issue - sometimes not fetching new tags when tag filtering enabled

analytically:
When using tag_filter or tag_regex without a branch specified, tags on commits not reachable from the default branch were not being detected.

Root cause: --single-branch clone and branch-specific fetch meant commits on divergent branches weren't fetched, so git rev-list --all couldn't find them and get_tags_matching_filter excluded the tags.

Fix: when tag filtering is enabled, clone without --single-branch and fetch all branch refs (refs/heads/*) so commits from all branches are available.

Fixes #448

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -117,7 +117,11 @@ export GIT_LFS_SKIP_SMUDGE=1
 
 if [ -d $destination ]; then
   cd $destination
-  git fetch origin $tagflag $branch -f
+  if [ "$tagflag" = "--tags" ]; then
+    git fetch origin 'refs/heads/*:refs/heads/*' --tags -f
+  else
+    git fetch origin $tagflag $branch -f
+  fi
   git reset --soft FETCH_HEAD
 else
   branchflag=""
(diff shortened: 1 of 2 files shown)


taylorsilva (approved):
LGTM - tested it out in a pipeline as well with a repo that has tags in the non-main branch. The resource was able to find them.

PR #472, opened 2026-04-27
Fix slow tag ancestors

C85297:
In #437 I added the `match_tag_ancestors` feature to return commits that were ancestors of matching tags.

That initial implementation turned out to be poorly optimised, and in large repos with many tags and many commits, a nested loop over each commit and each tag becomes very slow.

This pull request optimises the `match_tag_ancestors` feature:
1. Rather than resolving the commit each tag targets inside each loop iteration, each tag is resolved to its commit once, before the loop.
2. Check for commits that match one of the already resolved tag targets and instantly return
3. For all other commits, use `git tag --contains` to get a list of tags the commit is an ancestor of, and check if any of those tags match the eligible tags we're looking for.

My testing showed around 100x less subprocesses spawned and 40x faster check runs.

### DIFF
--- a/assets/check
+++ b/assets/check
@@ -260,16 +260,41 @@ get_tags_match_ancestors_filter() {
   # Sort commits so that we look at the oldest commits first
   local this_list_command="$list_command | git rev-list --stdin --date-order --first-parent --no-walk=unsorted --reverse"
   local list_output="$(set -f; eval "$this_list_command"; set +f)"
+
+  # Store all the tag names in an associative array for quick lookups.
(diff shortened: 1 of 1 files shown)


C85297:
It seems EasyCLA will require me to get a corporate CLA approval completed, this may take a while :/

C85297:
@taylorsilva Think the CLA should be sorted now :+1:

taylorsilva:
Awesome! Pulling in and I'll get it into a release.

PR #475, opened 2026-06-15
Sanitize invalid UTF-8 in git metadata

kurtmc:
It's possible for invalid UTF-8 characters to get into commit message so we are seeing the following error in production. I've added a sanitise step to elimite this issue.


kurtmc:
@taylorsilva I've reproduced the issue locally, and pushed a commit to my examples fork: https://github.com/kurtmc/examples/tree/utf8-emdash-repro

